PR #9, opened 2016-12-22
propose Autocrypt recommendations for replying to encrypted messages

dkg:
Replying to encrypted messages in the clear is a faux pas.  Autocrypt
should avoid doing this if it knows the keys for all recipients
targeted by the reply.

There are some sticky corner cases that the user might encounter when
replying to an encrypted mail if they do not already have the
necessary keys, unfortunately, but something along these lines is
probably necessary if we want this to be convenient and comprehensible
and safe for most users.

I welcome alternate proposals for how MUAs should deal with this
situation!

### DIFF
--- a/doc/level0.rst
+++ b/doc/level0.rst
@@ -421,11 +421,6 @@ during message composition:
    the message in cleartext, and default to encryption.  Prepare the
    message as an encrypted message.
 
-.. todo::
-
-   The Autocrypt recommendation should probably change depending on
-   whether the mail is a reply to an encrypted e-mail or not.
-
 Recommendations for single-recipient messages
 +++++++++++++++++++++++++++++++++++++++++++++
 
@@ -437,8 +432,9 @@ If the ``pah`` is ``null``, or if ``pah.key`` is known to be unusable
 for encryption (e.g. it is otherwise known to be revoked or expired),
 then the recommendation is ``disable``.
 
-If the ``pah`` is not ``null``, and ``prefer-encrypted`` is ``yes``,
-then the recommendation is ``encrypt``.
+If the ``pah`` is not ``null``, and ``prefer-encrypted`` is ``yes`` or
+the message being composed is a reply to an encrypted message, then
+the recommendation is ``encrypt``.
 
 If ``pah`` is not ``null``, and ``prefer-encrypted`` is either ``no``
 or ``nopreference``, then the recommendation is ``available``.
@@ -453,13 +449,36 @@ for each recipient individually.
 If any recipient has a recommendation of ``disable`` then the message
 recommendation is ``disable``.
 
-If every recipient other than "myself" (the e-mail address that the
+If the message being composed is a reply to an encrypted message, or
+if every recipient other than "myself" (the e-mail address that the
 message is ``From:``) has a recommendation of ``encrypt`` then the
 message recommendation is ``encrypt``.
 
 Otherwise, the message recommendation is ``available``.
 
+Cleartext replies to encrypted mail
++++++++++++++++++++++++++++++++++++
 
+As you can see above, in the common use case, a reply to an encrypted
+message will also be encrypted.  Due to Autocrypt's opportunistic
+approach, however, it's possible that ``pah`` is ``null`` for some
+recipient, which means the reply will be sent in the clear.
+
+To avoid leaking cleartext from the original encrypted message in this
+case, the MUA MAY prepare the cleartext reply without including any
+of the typically quoted and attributed text from the previous message.
+Additionally, the MUA MAY include brief text in message body along the
+lines of::
+
+  The message this is a reply to was sent encrypted, but this reply is
+  unencrypted because I don't yet know how to encrypt to
+  ``bob@example.com``.  If ``bob@example.com`` would reply here, my
+  future messages in this thread will be encrypted.
+
+The above recommendations are only "MAY" and not "SHOULD" or "MUST"
+because we want to accomodate a user-friendly level 0 MUA that stays
+silent and does not impede the user's ability to reply.  Opportunistic
+encryption means we can't guarantee encryption in every case.
 
 Encrypt outbound mail as requested
 ----------------------------------


doc/level0.rst
```diff
@@ -437,13 +433,31 @@ for each recipient individually.
...
+Cleartext replies to encrypted mail
++++++++++++++++++++++++++++++++++++
+
+As you can see above, in the common use case, a reply to an encrypted
+message will also be encrypted.  Due to Autocrypt's opportunistic
+approach, however, it's possible that ``pah`` is ``null`` for some
+recipient, which means the reply will be sent in the clear.
+
+To avoid leaking cleatext from the original encrypted message in this
+case, the MUA SHOULD prepare the cleartext reply without including any
```
hpk42: i'd say "MAY".

doc/level0.rst
```diff
@@ -437,13 +433,31 @@ for each recipient individually.
...
 
+Cleartext replies to encrypted mail
++++++++++++++++++++++++++++++++++++
+
+As you can see above, in the common use case, a reply to an encrypted
+message will also be encrypted.  Due to Autocrypt's opportunistic
+approach, however, it's possible that ``pah`` is ``null`` for some
+recipient, which means the reply will be sent in the clear.
+
+To avoid leaking cleatext from the original encrypted message in this
```
hpk42: typo at "cleaRtext"

hpk42:
Hum,  "reply to encrypted in the clear is a faux pas" is IMHO a bit tied to traditional mail encryption models of "securing the channel between two persons".  For opportunistic "replace cleartext" encryption it's not that clearly a faux-pas to not enable encryption.  Also consider a MUA that hides autocrypt/encryption mostly and a mailer who was unaware an encrypted mail was sent.  Why should we make replying hard then?  I know that K9/enigmail are currently quite explicit about showing encryption status but i also  think it's a good idea to reduce the clutter coming from crypto properties (too many traffic lights hurt ...).  

That being said, i think it's probably OK to just include the recommendation of "MAY" (see my inline comment) for now.  Also happy to hear what others add to the discussion.  Not completely settled myself on it.

dkg:
thanks for the feedback, @hpk42.  I agree with all of your points, and have updated the pull request to try to incorporate them.

josuttis:
To summarize what I wrote in the chat:
I think it all comes down to the general question: what is a useful reply to an encrypted email if I can't encrypt it. 
IMO, a reply to an encrypted email being automatically not encrypted is dangerous and suprising (because that's not what user will expect).
There should be at least SOME explicit confirmation necessary, whatever it is (warning about missing key, asking for confirmation, ...).
Also the general effect should be the same whenever we can't encrypt the reply (whether we have no key, the key exored, the pah is/got null, the preferecne is not to encrypt,...).  Of course, corresponding messages might differ to explain the problem.
In general, I see the following options:
- we recommend to encrypt (and let MUAs deal with handling the following errors)
- we recommend to MUAs to have some explicit interaction/confirmation
- we don't recommend anything and let the MUAs decide
IMO the trick with disabling quoting is too complicated.

hpk42:
@josuttis it's not clear what the user expects if we have opportunistic more invisible encryption.  the current "enigmail" model is not a 100% good reference point because it makes dealing with keys and encryption very visible and explicit to users.  

If you read https://tools.ietf.org/html/rfc7435#section-1.2 you'll find:

>  When less than complete protection is negotiated, there is no need to
   prompt the user with "your security may be degraded, please click OK"
   dialogs.  The negotiated protection is as good as can be expected.
   Even if not comprehensive, it is often better than the traditional
   outcome of either "no protection" or "communications failure".

RFC7435 is right in saying "Change of perspective" :) 

@dkg i do think adding a sentence that explains why we only say "MAY" would be good.  e.g. "Note that it's allowed for a level-0 MUA to stay silent and not impede the user's ability to reply. Opportunistic Encryption means we can't guarantee encryption." or something better.

dkg:
@hpk42 i've added the explanatory sentence that you suggested in the latest re-base of the dkg/reply-recommendation branch.  For future suggestions like this, feel free to push them to the branch directly!

dkg:
I'd like to resolve this, but we haven't seen a lot of commentary.  would someone else please review and either suggest changes or merge, or suggest an alternative?

doc/level0.rst
```diff
@@ -422,6 +417,7 @@ for encryption (e.g. it is otherwise known to be revoked or expired),
 then the recommendation is ``disable``.
 
 If the ``pah`` is not ``null``, and ``prefer-encrypted`` is ``yes``,
+or if the message being composed is a reply to an encrypted message,
```
npdoty: One could misread this as recommending ``encrypt`` when the message is a reply to an encrypted message, even if ``pah`` is ``null``, which won't work. It might be clearer to phrase this as:

> If the ``pah`` is not ``null``, and either ``prefer-encrypted`` is ``yes`` or the message being composed is a reply to an encrypted message,

npdoty:
(Not a core developer, just an interested observer.) I think opportunistic security -- where MUAs can decide to notify (or remove the quoted text) or simply send cleartext -- sounds like the right approach to me. 

Currently, PGP doesn't require that replies be sent encrypted, and there is no security policy in place (at least, in the implementations I've used) insisting that text that was sent encrypted always be sent encrypted. It's simply a common, polite convention not to include text that was encrypted in a cleartext reply, much the way it's a common convention not to expand the CC list if there's something sensitive in the quoted text, etc.

doc/level0.rst
dkg: thanks, i think i've adopted this cleanup suggestion in my latest revision.

hpk42:
Thanks @npdoty for the feedback and @dkg for the PR!

To be precise on this issue, I  think that whether to reply encrypted or not depends on if the other side explicitly choose encryption or just went with some invisible default.  If I explicitly sent encrypted i'd be happy if your MUA would pester you about sending a cleartext reply.  Otherwise i don't care.  

I don't think this information about the other side's user choice is precisely available although we might consider that if i know i gave out ``prefer-encrypt=no`` in my Autocrypt header and i receive an encrypted mail that it was an explicit choice.   Then again, i am fine for now with the merged PR.

dkg:
On Thu 2017-01-12 14:16:43 -0500, holger krekel wrote:

> To be precise on this issue, I think that whether to reply encrypted
> or not depends on if the other side explicitly choose encryption or
> just went with some invisible default.  If I explicitly sent encrypted
> i'd be happy if your MUA would pester you about sending a cleartext
> reply.  Otherwise i don't care.

do you think that means that if Alice manually clicks "encrypt", her MUA
should send that one e-mail with "prefer-encryped=true" regardless of
her internal preferences?

I'm not sure this makes sense if i try to simulate it through a
multi-message thread, even between two people, though (what about a
reply-to-an-encrypted-reply?).  But it might be worth thinking it
through.

It'd keep things simpler if we can just assume that the peer will reason
a certain way about a reply simply because it's a reply, i think.

  --dkg

hpk42:
On Fri, Jan 13, 2017 at 14:00 -0800, dkg wrote:
> On Thu 2017-01-12 14:16:43 -0500, holger krekel wrote:
> 
> > To be precise on this issue, I think that whether to reply encrypted
> > or not depends on if the other side explicitly choose encryption or
> > just went with some invisible default.  If I explicitly sent encrypted
> > i'd be happy if your MUA would pester you about sending a cleartext
> > reply.  Otherwise i don't care.
> 
> do you think that means that if Alice manually clicks "encrypt", her MUA
> should send that one e-mail with "prefer-encryped=true" regardless of
> her internal preferences?

No. I was talking about user choices, not what we currently have
on the wire. 

> I'm not sure this makes sense if i try to simulate it through a
> multi-message thread, even between two people, though (what about a
> reply-to-an-encrypted-reply?).  But it might be worth thinking it
> through.
> 
> It'd keep things simpler if we can just assume that the peer will reason
> a certain way about a reply simply because it's a reply, i think.

Sure, i just wanted to state my precise thought on the reply-encrypted
issue and that _if_ we want to evolve it further we might need some way
to know (on the wire) that the other side made an explicit choice.
I don't think it's neccessary to complicate the wire protocol, though
and so the case and PR is closed as is :)

PR #13, opened 2016-12-31
bunch of wording changes and clarifications

Valodim:
what it says on the tin

### DIFF
--- a/doc/level0.rst
+++ b/doc/level0.rst
@@ -57,16 +57,16 @@ decrypting.  It MUST be capable of assembling these keys into an
 OpenPGP certificate (RFC 4880 "Transferable Public Key") that
 indicates these capabilities.
 
-These secret keys MUST be protected from access by other applications
-or co-tenants of the device at least as well as the passwords the MUA
-retains for the user's IMAP or SMTP accounts.  These secret keys MUST
-never be sent over the wire to any other party.
+The secret key material is critical for security as in other end-to-end
+applications, and should be protected from access by other applications or
+co-tenants of the device, at least as well as the passwords the MUA retains for
+the user's IMAP or SMTP accounts.
 
-When an Autocrypt-enabled MUA configures an e-mail account, it should
-generate these keys and store them locally.  Then it should proceed to 
-try to "claim" the account to `lock out <lockout>`_ other MUAs of the
-same users.  In Level 0 only one MUA can send and receive encrypted
-mail through Autocrypt mechanisms.
+In Level 0 only one MUA can send and receive encrypted mail through Autocrypt
+mechanisms.  When an Autocrypt-enabled MUA configures an e-mail account, it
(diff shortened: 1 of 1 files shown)


doc/level0.rst
```diff
@@ -57,16 +57,16 @@ decrypting.  It MUST be capable of assembling these keys into an
...
 
-When an Autocrypt-enabled MUA configures an e-mail account, it should
-generate these keys and store them locally.  Then it should proceed to 
-try to "claim" the account to `lock out <lockout>`_ other MUAs of the
-same users.  In Level 0 only one MUA can send and receive encrypted
-mail through Autocrypt mechanisms.
+In Level 0 only one MUA can send and receive encrypted mail through Autocrypt
+mechanisms.  When an Autocrypt-enabled MUA configures an e-mail account, it
+first tries to "claim" the account, to `lock out <lockout>`_ other MUAs of the
+same users.  If the claim was successful, the MUA should proceed to generate
```
dkg: I actually think a responsible MUA should generate provisional keys *before* attempting to claim the account, to minimize the race window -- that way, if the account appears to be unclaimed, it can claim it immediately.

doc/level0.rst
```diff
@@ -92,25 +92,21 @@ announcement and places it in the special location.
...
+material for, it SHOULD include an Autocrypt header. This header contains the
+associated public key material as ``key=`` attribute, and the same sender
+address that is used in the ``From`` header in the ``to=`` attribute to confirm
+the association. The most minimal Level 0 MUA will only include these two
 attributes.
 
-If the ``From:`` address changes during message composition (e.g. if
-the user selects a different outbound identity, the Autocrypt-capable
-client MUST change the ``Autocrypt:`` header.
-
```
dkg: why was this paragraph removed?

doc/level0.rst
```diff
@@ -119,62 +115,70 @@ the following sections for header format definitions and parsing.
...
-attribute name starts with an underscore (``_``), this is a
-"non-critical" attribute.  The MUA MUST ignore the non-critical
-attribute and continue parsing the header.  If the unknown attribute
-name does *not* start with an underscore, it is a "critical"
-attribute, and the MUA must skip the header as invalid.
+Additional attributes unspecified here are also possible before the ``key``
+attribute.  If an attribute name starts with an underscore (``_``), it is a
+"non-critical" attribute.  The MUA SHOULD ignore non-critical attributes, but
+MUST treat the entire header as invalid if it encounters a "critical" attribute
+it doesn't support.
```
dkg: This paragraph seems weaker than the original: the original is about encountering unknown attributes, but the replacement seems to suggest that MUAs SHOULD ignore non-critical attributes it is aware of.

dkg:
I've merged this and added a few fixes on top of it to reflect the comments above.

I also added a .editorconfig file so that we can avoid linewrapping clashes :)

PR #15, opened 2017-01-03
add sequence diagrams for the examples.

juga0:
* add plantuml source files (*.puml) and images (.svg, .png)
* if plantuml sphinx extension is installed, the plantuml code can be include in the sources with the `. uml:: ` directive.


doc/examples.rst
```diff
@@ -52,7 +52,9 @@ Establishing encryption happens as a side effect when people send each other mai
 "Happy path" example: 1:1 communication
 ---------------------------------------
 
-.. image:: images/autocrypthappy.svg
+.. image:: ./images/autocrypthappy.svg
```
dkg: why do we need the ./ prefix here?

juga0: not neede

dkg:
I've just pushed the branch features/diagrams-v2 to the origin, which is a squashed rebase of the proposal here.

I like the documentation it produces, but to use it we need manual installation on debian of two packages:

 * plantuml
 * python-sphinxcontrib.plantuml

(i've filed a bug with debian about the missing dependency from python-sphinxcontrib.plantuml to plantuml but it hasn't been assigned a bug number yet)

juga0, can you take a look at features/diagrams-v2 and see if it meets your expectations?

hpk42:
do you mean installation of packages needs to happen on readthedocs?

juga0:
yes, maybe it'll be included https://github.com/rtfd/readthedocs-docker-images/pull/17.

Same as with any SVG image, inkscape is a dependency. There's a similar issue for sphinx to include SVG in latex (https://github.com/sphinx-doc/sphinx/issues/1907). Currently they are converted via Makefile: https://github.com/autocrypt/autocrypt/pull/14#issue-198468872.

The alternative is to upload the generated SVG/PDF images, as commented here https://github.com/autocrypt/autocrypt/pull/15#issue-198475150, or even only PDF (or PNG) to don't depend also on inkscape.

dkg:
On Thu 2017-01-12 09:46:02 -0500, juga0 wrote:
> The alternative is to upload the generated SVG/PDF images, as
> commented here
> https://github.com/autocrypt/autocrypt/pull/15#issue-198475150, or
> even only PDF (or PNG) to don't depend also on inkscape.

this alternative produces sub-optimal docs, though, because the
converted PNG images don't contain searchable text the way the SVG
does.  (it also means committing generated files into the repo, which is
i could live with, but seems untidy if we don't need to do it)

    --dkg

juga0:
Plantuml seems to be already in the docker image that readthedocs.org uses (https://github.com/rtfd/readthedocs.org/issues/407).
But anyway, these diagrams doesn't add any relevant information, and they just add "more pages". So i'm ok closing this PR.

PR #21, opened 2017-01-12
fix #18 by clarifying we are working from useaibility, incremental de…

### LINKED ISSUE #18: Clarify that "level 0" is about guaranteeing a user experience, not the smallest possible interesting featureset
I've heard from some MUA developers that they felt blind-sided by the "level 0" terminology.  they were thinking that "level 0" would be something like the smallest possible feature set to be interesting to a developer.

Instead, i think the autocrypt project sees "level 0" to be the smallest possible feature set to be interesting to a *user*, while it contains several different moving pieces that a developer might want to consider individually (and might individually provide some actual value in an MUA targeted toward more-sophisticated users) in the classic "divide and conquer" approach to implementation.

I don't think we should change the "level 0" concept -- i think it's important that autocrypt stays user-focused.  But it'd be nice to communicate this better to developers so that they don't feel confused or surprised by the change of focus.

hpk42:
…ployment

perspectives.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -37,14 +37,15 @@ The technical Autocrypt approach
 --------------------------------------
 
 Autocrypt uses regular e-mail messages between people to piggyback
-necessary information to allow encrypting subsequent messages.  
-Under the hood, Autocrypt uses e-mail headers for this information
-transfer. By default, no key management is visible to users.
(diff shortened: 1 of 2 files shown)


dkg:
This looks good to me, modulo a few spelling/grammar cleanups.  I'll merge it and then clean those up.

PR #28, opened 2017-01-23
Bot takes Alice´s address from the from attribute

titusz:


titusz:
Ok it seems i got this wrong.... the to-attribute is meant to refer to the Autocrypt header to-attribute?  Isn´t that redundant with the mail header from attribute?

Valodim:
The from header can change, e.g. in mailing lists that forward in a dmarc compatible ways

titusz:
Well yes that makes sense ... In any case malicious user could advertise wrong autoencrypt headers in the name of others .... but i guess there is not much that can be done about that.

Valodim:
They can, but such troubles should be mitigated by spam filters. And as a targeted attack, it's not exactly covert :)

PR #33, opened 2017-02-08
Send autocrypt

juga0:
Not needed, but felt like to play :-)


hpk42:
thanks, @juga0.  Note that for code to enter the actual autocrypt bot code base i'll usually aim for tests as it otherwise becomes gambling if merging PRs, writing new code breaks things.

PR #35, opened 2017-02-09
Minor Cleanup, renaming of code in gpg.py 

hpk42:


dkg:
These changes look technically reasonable from what i can see, but i don't understand why the rename is necessary.

What does "BinGPG" mean?

I plan to merge them because it sounds like they'll facilitate other work, but a clearer commit message would be welcome.

hpk42:
BinGPG because we are using the gpg binary and not e.g. "gpgme". We can also call it "gpgbin" FWIW :)

dkg:
hm, ok.  i guess i'm biased because i know that gpgme itself actually just calls the gpg binary under the hood anyway, so i don't see much of a distinction there :/

i'd have called it "gpgwrapper" or something like that, if you wanted to make this distinction clearer, but i'm fine with it as it stands for now.

PR #42, opened 2017-03-06
Rework level-0 state to implement soft resets

BjarniRunar:
See individual commits for more details on what changes I'm making to the spec.

edit: This is part of the output from the AutoCrypt pre-IFF workshop.

### DIFF
--- a/doc/level0.rst
+++ b/doc/level0.rst
@@ -238,8 +238,34 @@ Autocrypt-compatible agents SHOULD track and store in
 necessarily the literal header emitted (for the literal header, see
 next section).  The ``pah`` MUST contain the following fields:
 
- * ``key`` -- the raw key material, after base64 decoding
- * ``prefer_encrypted`` -- a tri-state: ``nopreference``, ``yes``, or ``no``
+ * ``key``: the raw key material, after base64 decoding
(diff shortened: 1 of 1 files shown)


Valodim:
Just from a quick glance, I find "pah" to be a confusing term to use next to "message_pah". Maybe go for "stored_pah"?

doc/level0.rst
```diff
@@ -238,8 +238,34 @@ Autocrypt-compatible agents SHOULD track and store in
...
+ * ``key``: the raw key material, after base64 decoding
+ * ``prefer_encrypted``: a quad-state: ``nopreference``, ``yes``, ``no`` or ``reset``
+
+Optionally, an agent MAY store and maintain the following data in
+order to provide clearer feedback to the user:
+
+ * ``counting_since``: The UTC timestamp of when we started counting
+ * ``count_have_ach``: A count of parsed AutoCrypt headers
+ * ``count_no_ach``: A count of messages without AutoCrypt headers
+ * ``bad_user_agent``: The apparent user-agent (if known) of the last
```
dkg: How is a mail user agent expected to accumulate these counts, given that it might see the same message multiple times?  From the algorithms described here, i don't see a way to avoid double-counting when the MUA happens to look at a message twice.

PR #43, opened 2017-03-09
No more SMA

Valodim:
I started removing the SMA parts of the spec, and replacing them with the "check in Sent box" and "secret key transfer via self-sent msg" mechanisms.

There are a couple of things that are non-obvious for the message format. I'll get back to this, but wanted to push what I have as a starting point

### DIFF
--- a/doc/level0.rst
+++ b/doc/level0.rst
@@ -99,7 +99,16 @@ this mechanism, to make sure that secret keys are never stuck on the
 device they were generated on, but can be retrieved by other Autocrypt
 clients in a predictable way.
 
-[technical explanation]
+- create self-sent message
+- contains secret key
+- symmetric encryption format
+- high-entropy passphrase
+  - bit strength?
+  - is s2k sufficient for a kdf? or scrypt on top?
+  - must not end up in the user's mail store!!
+- message format
+  - attachment? something else?
+- message text
 
 Header injection in outbound mail
 ---------------------------------
--- a/doc/level0.rst
+++ b/doc/level0.rst
@@ -34,23 +34,19 @@ has credentials and capabilities to perform these network services:
   MUA can control the entire message being sent, including both
   message headers and message body.
 
-- The ability to receive e-mail where the MUA gets access to
-  the entire message being received, including both message
-  headers and message body.
+- The ability to receive e-mail where the MUA gets access to the
+  entire message being received, including both message headers and
+  message body.
 
-- Access to a special (IMAP) Shared Message Archive (SMA) folder which
-  can be accessed by all MUAs of a user's devices to co-ordinate
-  between them.  In Level 0 this is only used for :ref:`ensuring that only
-  one MUA has Autocrypt enabled for an e-mail account at once
-  <lockout>`.
+- Optionally, a way to scan the user's Sent folder for mail with
+  specific headers.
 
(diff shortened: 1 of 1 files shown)


doc/level0.rst
```diff
@@ -71,40 +67,48 @@ end-to-end applications, and should be protected from access by other
...
+to behavior that is unpredictable and confusing for the user.
+
+As a simple measure of mitigation, Level 0 MUAs SHOULD check before
+key generation whether there is evidence in the user's mailbox of
+other active Autocrypt clients. To do this, they SHOULD scan the
+user's Sent folder for mail that contains Autocrypt headers. If such
+mail exists, the MUA SHOULD warn the user and abort key generation,
+unless explicitly instructed to proceed regardless.
+
+In cases where this mechanism cannot be provided, the MUA MUST inform
```
dkg: I'm not sure how an implementor is going to be able to tell whether "this cannot be provided".  What is "this" exactly?  If it's "i don't know how to check the user's Sent Mail" is that a function of the MUA itself?  or its ability to access a particular mailbox?  or a property of the mailbox, regardless of how the MUA knows how to access it?

This seems like something that would be hard for an implementer to reason about without clearer guidance.

doc/level0.rst
```diff
@@ -71,40 +67,48 @@ end-to-end applications, and should be protected from access by other
...
+the user that they should only set up one Autocrypt client.
+
+To solve this problem in a better way, bi-directional communication
+between the user's different MUAs is required. However, this is out of
+scope for Level 0.
+
+Secret Key Transfer
```
dkg: Why include this bit in this pull request?  I think it'll be easier to sort out the message (whether it's backup or transfer or however we're framing it) as a distinct discussion.

doc/level0.rst
Valodim: "this" refers to the mitigation mechanism from the preceding paragraph, for cases where the implementation doesn't (or can't) honor the SHOULD. in those cases, it MUST inform the user they should only set up one autocrypt client. would it be clearer to say "this mitigation mechanism"? or something else?

doc/level0.rst
Valodim: I wanted to note down at least a rough roadmap for this, in order to make it easier to pick up by me or someone else who wanted to flesh out the specifics. similar to other TODOs we have in the docs.

dkg: sure, i think it's fine as a rough roadmap, though the choice of language ("transfer" instead of "backup") already appears to make some explicit decisions.

doc/level0.rst
dkg: The problem here lies in helping implementors know specifically when they can and cannot implement the mitigation mechanism, which i think is ill-defined.  I might go ahead and merge this PR and try to take a stab at defining it, if that's useful.

Valodim: Sounds good, go ahead!

doc/level0.rst
Valodim: I'm not clear on the specifics anymore, but I think that name was what we used at Valencia to distinguish it from a mechanism meant to restore at a *later* point.

doc/level0.rst
dkg: this clarification is now PR #50

PR #47, opened 2017-04-12
feature: reply encrypted to encrypted mails

azul:
... and unencrypted to unencrypted mails as long as the peer did not
set prefer encrypt.

### DIFF
--- a/ui/config/autocrypt-ui.js
+++ b/ui/config/autocrypt-ui.js
@@ -301,6 +301,7 @@ reply_to_msg = function(msg) {
     ui['body'].value = indent(msg['body']);
     replying_to = msg;
     pane('compose');
+    ui['encrypted'].checked = ui['encrypted'].checked || msg['encrypted'];
 };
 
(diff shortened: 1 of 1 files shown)


ui/config/autocrypt-ui.js
```diff
@@ -301,6 +301,7 @@ reply_to_msg = function(msg) {
     ui['body'].value = indent(msg['body']);
     replying_to = msg;
     pane('compose');
+    ui['encrypted'].checked = ui['encrypted'].checked || msg['encrypted'];
```
Valodim: doesn't this lead to problems, e.g. in cases where not all keys are available?

azul: Hi @Valodim,
So far this is only Alice talking to Bob and autocrypt only... so for that scenario we will always have our peers key when replying.
But the point you are bringing up is still valid at least for multi recipient mails and reply all. I think dropping encryption in that case should result in a warning to the user.

PR #54, opened 2017-06-06
specify key transfer format

Valodim:
This adds a document describing (my understanding of) @BjarniRunar's transfer format idea. I don't think it's necessary to make this part of the spec proper, so the idea is to have a standalone document outlining this idea, and referring to it as recommendation from the main doc.

This is working draft to have a starting point. Statements about file handling or usability improvements are true are speculation, since I have not run any tests (user or otherwise) to verify.

Comments & critique welcome. I'm also a newbie at ReST, so improvements in that regard are also welcome.

### DIFF
--- a/doc/transfer-format.rst
+++ b/doc/transfer-format.rst
@@ -1,0 +1,109 @@
+Key Transfer Format
+===================
+
+This document describes a format recommendation for secret key backup
+attchments. The Autocrypt Key Transfer Format is designed to work as an
+attachment in a mail client that can be handled automatically by Autocrypt
+clients, but also strives for usability when saved as a file in agnostic
+clients.
+
+This goal is achieved by two mechanisms:
+
+- The MIME part MUST have a Content-Disposition of "attachment". Its filename,
+  as specified in Content-Type and Content-Disposition header parameters, MUST
+  have an extension of ".html".
+- The file MUST be formatted as proper HTML. It SHOULD give a human-readable
+  description at the beginning, and MUST include the encrypted key as a
+  contiguous block of ASCII-armored data, with no indentation or HTML tags
+  inside. This block SHOULD reside inside a <pre> block for readability.
+
+With this technique, the way the attachment is handled when saved to disk by the
+user is independent from its handling by clients aware of its Content-Type. HTML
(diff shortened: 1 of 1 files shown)


doc/transfer-format.rst
```diff
@@ -0,0 +1,109 @@
...
+
+An exemplary secret key backup MIME part would look as such:
+
+.. code-block:: html
+
+  Content-Type: application/pgp-key-backup; name="Autocrypt-Key for name-at-example.org.html"
+  Content-Description: OpenPGP Digital Signature
+  Content-Disposition: attachment; filename="Autocrypt-Key for name-at-example.org.html"
+
+  <html><body><h1>Autocrypt Key</h1><p>
```
hpk42: Instead of talking about "key" here and below i suggest to rather talk about "Autocrypt secret key number" or so.  Basically a term that makes some sense for someone who has never heart of "keys" and who may even use some autocrypt-by-default client (so even less context to be assumed).

Valodim: While the idea to make the headline or description easier to understand is of course good, I have my doubts about this particular suggestion. Both terms "secret key" and "number" (in this context) belong on the "never use except for crypto nerds" list, imo.

BjarniRunar:
Discussions about how to phrase the helpful text aside (always useful and worth spending time on), I think you've accurately captured both the intent and the mechanism.

Good work, thanks for taking the time to turn my brain dump into a useful document.

Valodim:
Thanks :)

One more potential todo in this document is adapting the `pgp-key-backup` content-type to whatever actual type we choose.

doc/transfer-format.rst
hpk42: Hum, not sure i agree.  signal uses "secret numbers" IIRC and people are used to this through PINS/PUKS etc.  However, my comments here do not need to hold up the merging of this PR as i agree with Bjarni that it looks good basically!

Valodim: do you mean "safety numbers"? they do, but that's when there are actually a bunch of numbers shown on screen with a clear and simple description of what to do with them (compare).

PR #55, opened 2017-06-08
specify transfer mechanism

Valodim:
First attempt at specifying a transfer mechanism. Most of this is super simple, but it turned out to be quite a lot of text regardless. :hurtrealbad:

There are a couple of TODOs left and points for discussion, but we should be able to work from here.


azul:
Read it once. Looks good to me. I have not looked into the crypto details but at first sight they make sense.
I find the mixture of terms referring to transfer (export, key transfer mechanism) and backup (backup code) confusing. First term 'backup code' is mentioned i was not sure what it was referring to.

doc/level0.rst
```diff
@@ -92,26 +92,150 @@ To solve this problem in a better way, bi-directional communication
...
+
+Setup Code
++++++
+
+The setup code MUST be generated by the implementation itself using a
+CSPRNG, and presented directly to the user for safekeeping. It MUST
+NOT be included in the Autocrypt Setup Message, or otherwise
+transmitted over e-mail. Its format SHOULD be 24 uppercase
+alphanumeric characters, divided into six blocks of four, separated by
+dashes. The dashes are part of the secret code. This format holds 124
```
hpk42: some of us think it's better to use simply numbers which would amount to roughly 9 * 4 number blocks.   Some user studies (ask Dominik/vincent) indicate that using numbers produces the least confusion/errors when they are asked to type it in.  Also avoiding latin characters probably helps with chinese and other non-latin based alphabets. 

Maybe we can also use an error correction number block so that we have 10 * 4 numbers and a user can mistype 2 or 3 numbers.

hpk42: also, it probably makes sense for interop between autocrypt clients to fix the setup code format also in term of length.  Therefore "MUST" instead of "SHOULD".

doc/level0.rst
```diff
@@ -92,26 +92,150 @@ To solve this problem in a better way, bi-directional communication
...
+bits of data, it is designed to be unambiguous, pronounceable, and
+split into blocks that can be easily kept in short term memory. For
+instance:
+
+``AB1D-E2GH-IJK3-4NOP-Q5ST-XYZ6``
+
+When this form of Setup Code is used, the ASCII-armored block in the
+payload SHOULD include an OpenPGP Armor header called
+"Autocrypt-Setup-Message" with a value of "v1", to indicate to
+receiving implementations the format of the passphrase.  This is a
```
hpk42: rename "passphrase" to "setup code".

doc/level0.rst
```diff
@@ -92,26 +92,150 @@ To solve this problem in a better way, bi-directional communication
...
+useful indicator at message import time.
+
+Setup Message Creation
+++++++
+
+An Autocrypt client MUST NOT create an Autocrypt Setup Message without
+explicit user interaction.  When the user takes this action for a
+specific account, the client:
```
hpk42: patrick, bjoern, oliver, holger think it's worthwhile to allow for automatic recognition of an incoming setup message from another device.  it's probably helpful to have an ``Autocrypt-Setup`` header which allows for trivial and more precise recognition of setup messages (otherwise we might need to look at all bodies of all mails which have ``From: X To: X`` settings).  Also it's possible that a user has different identities/aliases configured on the sending and the receiving device.

doc/level0.rst
```diff
@@ -92,26 +92,150 @@ To solve this problem in a better way, bi-directional communication
...
+
+ * The client should prompt the user for their corresponding Setup
+   Code.  If there is an ``Autocrypt-Setup-Message: v1`` OpenPGP
+   header, the client may choose to present the user with a
+   specialized input dialog that better assists the user with input in
+   this particular format.
+ * The client should try decrypting the message with the supplied
+   Setup Code.  If it decrypts:
+ * The client should verify that the User ID on the key matches the
+   User ID on the relevant account.
```
hpk42: We further up say that the User ID is only "decorative".  As there are also aliases/multiple identities to the same acocunt possibly, it's better to not require this kind of match.  As we already have a matching "Setup Code" which successfully decrypts the setup message we can rather assume that the key is for the "X" address in ``For: X  To: X`` like we usually do when parsing autocrypt headers.

hpk42: One problem here is that a naive implementation of e.g. asking gpg to decrypt a setup message is that it might be encrypted not through the symmetric setup code but asymetrically to a possibly pre-existing key on the second device.  Care needs to be taken (e.g. for enigmail) to make sure the setup message is actually using symmetric encryption.

doc/level0.rst
```diff
@@ -92,26 +92,150 @@ To solve this problem in a better way, bi-directional communication
...
+
+Why were some of these choices made?
+++++
+
+We chose salted+iterated S2K.  While the use of a memory-hard KDF like
+scrypt or argon2 would be desirable in the future, this is not
+specified in OpenPGP so far, and it is a bigger concern to preserve
+compatibility and avoid friction with presently deployed OpenPGP
+software.
+
```
hpk42: What is missing here is a description of the algorithm chosen and the precise pgp message (like it is done with type=p keys).

hpk42:
merging already because it's a big improvement.  some issues remain like commented in the PR here.

PR #58, opened 2017-06-10
move to "prefer-encrypt=mutual"

Valodim:


hpk42:
i like the PR -- i suggest to add a little paragraph explaining the movtivation of "prefer-encrypt=mutual"  before the wire-protocol is discussed in detail.

PR #59, opened 2017-06-10
use addr as the email address attribute

azul:
Also always refer to attributes without the trailing `=`.

addr is a bit more specific (grep is easier now).


hpk42:
everybody cheers!

PR #63, opened 2017-06-10
Put secret key into a mime structure, add prefer-encrypt into its hea…

r10s:
…der.


doc/level0.rst
```diff
@@ -133,8 +133,9 @@ specific format, which contains a payload protected by the setup code.
   symmetrically encrypted data, and MAY include other text above or
   below the ASCII-armored data, which MUST be ignored while
   processing.
-- Decrypting the payload should produce an ASCII-armored OpenPGP
-  transferable secret key.
+- Decrypting the payload MUST produce a multipart/mixed mime structure
+  with an optional Autocrypt:-header containing the user's prefer-encrypt state;
+  one part is an ASCII-armored OpenPGP transferable secret key.
```
hpk42: If it'd be a literal ``Autocrypt`` header we would need to distinguish it from "normal" Autocrypt header -- if a header, then maybe better use a ``Autocrypt-Settings`` or ``Autocrypt-OwnState`` header to not have to distinguish it from the "normal" one?

doc/level0.rst
```diff
@@ -237,6 +238,57 @@ While the message structure is complex, it's actually fairly easy to
...
+	</body>
+	</html>
+	---break---
+
+The encrypted messages contains:
+
+::
+
+	Content-type: multipart/mixed; boundary="---break2---"
+	Autocrypt: prefer-encrypt=mutual
```
hpk42: Again, ``Autocrypt-Settings`` or so looks better.

hpk42:
included in #75

PR #66, opened 2017-06-11
(proxying for dkg) Clarify MUSTs/SHOULDs around Setup Code

hpk42:
We want to require a Level 0 Autocrypt client to generate setup codes
using the standard form.

However, we understand that future versions of Autocrypt might change
the expected form (e.g. if there is any research on better
high-entropy transfer mechanisms).

We want Level 0 clients to be able to import Autocrypt Setup Messages
generated with those forms as well.

So on export, L0 MUST generate setup codes in our way, and MUST
include the "Autocrypt-Setup-Message: v1" header in the OpenPGP block
to indicate that it has done so.  And all clients MUST NOT include the
the header if they don't follow this form.

And on import, L0 MAY provide a specialized input dialog if the header
is present, but MUST provide straight UTF-8 string input as a fallback
if it encounters an Autocrypt Setup Message which doesn't have the
header.


doc/level0.rst
```diff
@@ -137,32 +137,37 @@ specific format, which contains a payload protected by the setup code.
...
+of 24 uppercase alphanumeric characters, divided into six blocks of
+four, separated by dashes. The dashes are part of the secret
+code. This format holds about 124 bits of data. It is designed to be
+unambiguous, pronounceable, and split into blocks that can be easily
+kept in short term memory. For instance::
+
+    AB1D-E2GH-IJK3-4NOP-Q5ST-XYZ6
+
+An Autocrypt Setup Message payload that uses this structure for its
+setup code SHOULD include the following header in the outer OpenPGP
```
r10s: Requiring an Autocrypt-Setup-Message header with some value would allow using this header to identify the Setup Message. The SHOULD in here refers to the value of the header.

doc/transfer-format.rst
```diff
@@ -28,82 +28,83 @@ still easily be processed by OpenPGP tools. In particular, GnuPG ignores the
...
+Setup Code ``D9VN-RD7A-7T3B-BWG7-LEPY-3NYT``.
 
 .. code-block:: html
 
   Content-Type: application/pgp-key-backup; name="Autocrypt-Key for name-at-example.org.html"
   Content-Description: OpenPGP Digital Signature
   Content-Disposition: attachment; filename="Autocrypt-Key for name-at-example.org.html"
 
-  <html><body><h1>Autocrypt Key</h1><p>
+  <html><body><h1>Autocrypt Key for <tt>example.org</tt></h1><p>
```
r10s: this should be an email address, right?

doc/transfer-format.rst
```diff
@@ -28,82 +28,83 @@ still easily be processed by OpenPGP tools. In particular, GnuPG ignores the
...
 
   This is an Autocrypt Key file. When imported into an Autocrypt capable E-Mail
   client with its backup code, it can be used to read encrypted E-Mail sent to
-  name@example.org.
+  <tt>name@example.org</tt>.
 
   </p><pre>
   -----BEGIN PGP MESSAGE-----
+  Autocrypt-Setup-Message: v1
```
r10s: If this was in a mail header it would allow identifying the email and rely on existing header parsing rather than requiring special treatment.

doc/transfer-format.rst
r10s: Autocrypt Setup Message created by mailclient.org for email address?

doc/level0.rst
azul: discussed in review of #63

doc/transfer-format.rst
azul: discussed in review of #63

PR #69, opened 2017-06-11
Define "effective date" before using it

dkg:


doc/level0.rst
```diff
@@ -390,10 +394,6 @@ times, such as upon receipt or display. When this happens, the
 Autocrypt state for the sending peer is updated with this new
 information. This update process depends on:
```
Valodim: the update process still depends on the effective date. deleting the effective date here paragraph leaves the list incomplete in regard to this statement.

PR #85, opened 2017-06-11
Move input format back into OpenPGP armor header

### LINKED ISSUE #80: ASM: Move Input-Type to an OpenPGP header
@Valodim proposed this because it is not autocrypt-specific, and it can be handed off to the OpenPGP submodule.

azul:
We want to hand this to the part of the client that is responsible
for decryption and displaying the key entry. This may not be the
same part that parses the message headers.

fixes #80

### DIFF
--- a/doc/level0.rst
+++ b/doc/level0.rst
@@ -174,8 +174,8 @@ in the outer OpenPGP armor::
 
     Passphrase-Format: alphanumeric
 
-This attribute MUST NOT be present if the Setup Code does
-not match the format described above.
+This OpenPGP header MUST NOT be present with this value if the Setup
+Code does not match the format described above.
 
 Setup Message Creation
 ++++++++++++++++++++++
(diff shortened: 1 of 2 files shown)


doc/level0.rst
```diff
@@ -170,10 +169,10 @@ kept in short term memory. For instance::
...
-setup code SHOULD include the following ``input-type`` attribute in
-the top-level ``Autocrypt-Setup-Message`` header::
+setup code SHOULD include the following ``Passphrase-Format`` header
+in the outer OpenPGP armor::
 
-    Autocrypt-Setup-Message: v0; input-type=alphanumeric
+    Passphrase-Format: alphanumeric
 
 This attribute MUST NOT be present if the Setup Code does
 not match the format described above.
```
dkg: This is a header and not an attribute.

PR #91, opened 2017-06-11
Proposing numeric input instead of alphanumeric input

r10s:


dkg:
As the person leaning toward zbase32, I'm fine with this change.

I really like the 3×3 grid representation of the blocks, which i think makes it easier to keep track of where the user is in the input sequence.

PR #97, opened 2017-06-13
recommend against password protecting keys

### LINKED ISSUE #96: recommend against passwords for the OpenPGP key
We talked about this a bunch but have not included it in the spec yet.

azul:
(with @Valodim)

fixes #96

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -417,6 +417,12 @@ public keys.  It MAY support other OpenPGP key formats found in
 a ``type=1`` Autocrypt header (for example, by passing it agnostically
 to an OpenPGP backend for handling).
 
+We recommend against protecting the private key with a password. All
+encrypted outgoing messages MUST also be signed, which would require the
+user to enter their password for both reading and sending mail. This
+introduces too much friction to become part of a routine daily workflow.
+Protection of the user's keys at rest is achieved more easily and
+securely with full-disk encryption.
 
 Internal state storage
 ----------------------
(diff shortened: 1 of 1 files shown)


Valodim:
:+1:

doc/level1.rst
```diff
@@ -417,6 +417,12 @@ public keys.  It MAY support other OpenPGP key formats found in
 a ``type=1`` Autocrypt header (for example, by passing it agnostically
 to an OpenPGP backend for handling).
 
+We recommend against protecting the private key with a password. All
```
hpk42: i'd like to be specific: "MUAs SHOULD NOT protect the secret Autocrypt key with a password."

doc/level1.rst
```diff
@@ -417,6 +417,12 @@ public keys.  It MAY support other OpenPGP key formats found in
...
+encrypted outgoing messages MUST also be signed, which would require the
+user to enter their password for both reading and sending mail. This
+introduces too much friction to become part of a routine daily workflow.
+Protection of the user's keys at rest is achieved more easily and
+securely with full-disk encryption.
```
hpk42: i'd add ", which also protects all kinds of other user data. Moreover, some architectures (e.g. mobile) already protect from other programs reading a mail app's data".

doc/level1.rst
hpk42: maybe add a subsection title of "passphrase-less secret keys" to make this something we can link to easily.  I consider this a major usability issue, not just a recommendation on the side.

doc/level1.rst
Valodim: I wouldn't add the second part. While true, it weakens our point for architectures that *don't* have that property. I like the first idea though, how about just adding "and other files"?:

"Protection of the user's keys and other files at rest is achieved more easily and securely with full-disk encryption."

PR #107, opened 2017-06-13
Add an option to display the first two characters of the setup code

r10s:
An idea how to display the first two characters of the setup code as discussed last weekend.


Valodim:
Nice! We should probably mention the loss of entropy as part of the motivation

PR #109, opened 2017-06-13
Alternative Passphrase-Begin suggestion

Valodim:
Alternative suggestion based on #107 by @r10s. I changed the "12" to "95" to have less generic numers, moved the wording around a bit and merged the two examples into one.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -161,42 +161,40 @@ NOT be included in the cleartext of the Autocrypt Setup Message, or
 otherwise transmitted over e-mail.
 
 An Autocrypt Level 1 client MUST generate a setup code as UTF-8 string
-of 36 numeric characters, divided into nine blocks of
-four, separated by dashes. The dashes are part of the secret
-code and there are no spaces. This format holds about 119 bits of entropy. It is designed to be
-unambiguous, pronounceable, script-independent (chinese, cyrillic etc.),
-easily input on a mobile device and split into blocks that are easily
-kept in short term memory. For instance::
+of 36 numeric characters, divided into nine blocks of four, separated
+by dashes. The dashes are part of the secret code and there are no
+spaces. This format holds about 119 bits of entropy. It is designed to
+be unambiguous, pronounceable, script-independent (chinese, cyrillic
+etc.), easily input on a mobile device and split into blocks that are
+easily kept in short term memory. For instance::
 
-    1203-1923-2307-
+    9503-1923-2307-
     1980-7833-0983-
     1998-7562-1111
 
 An Autocrypt Setup Message payload that uses this structure for its
-setup code SHOULD include the following ``Passphrase-Format`` header
-in the outer OpenPGP armor::
+setup code SHOULD include a ``Passphrase-Format`` header with value
+``numeric9x4`` in the ASCII-armored data. This allows providing a
+specialized input form during decryption, with greatly improved
+usability.
 
-    Passphrase-Format: numeric9x4
+As a further measure to improve usability, it is RECOMMENDED to reveal
+the first two digits of the first block in a ``Passphrase-Begin``
+header, sacrificing about 7 bits of entropy. Those digits can be
+pre-filled during decryption, which reassures the user that they have
+the correct code before typing the full 36 digits. It also helps
+mitigate a possible type of phishing attack that asks the user to
+input their setup code.
 
-This OpenPGP header MUST NOT be present with this value if the Setup
-Code does not match the format described above.
+The headers might look like this::
 
-Note: if the autocrypt implentation makes the passphrase easier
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -161,42 +161,40 @@ NOT be included in the cleartext of the Autocrypt Setup Message, or
...
-Code does not match the format described above.
+The headers might look like this::
 
-Note: if the autocrypt implentation makes the passphrase easier
-to read by inserting spaces around the dashes, then the implemenation
-should warn the user that spaces are not part of the passphrase, because
-not all OpenPGP implementations make it easy for the autocrypt implementation
-to filter the passphrase.  (GnuPG, for instance, uses pinentry to retreive the passphrase, which
-means the autocrypt implementation cannot check the passphrase's format.)
+    Passphrase-Begin: numeric9x4
```
r10s: The first line should read `Passphrase-Format: numeric9x4`

doc/level1.rst
```diff
@@ -161,42 +161,40 @@ NOT be included in the cleartext of the Autocrypt Setup Message, or
...
-This OpenPGP header MUST NOT be present with this value if the Setup
-Code does not match the format described above.
+The headers might look like this::
 
-Note: if the autocrypt implentation makes the passphrase easier
-to read by inserting spaces around the dashes, then the implemenation
-should warn the user that spaces are not part of the passphrase, because
-not all OpenPGP implementations make it easy for the autocrypt implementation
-to filter the passphrase.  (GnuPG, for instance, uses pinentry to retreive the passphrase, which
-means the autocrypt implementation cannot check the passphrase's format.)
```
r10s: I also though over this idea and did not like the general recommendation to the user to be warned not to enter spaces:
- if the GUI does not allow the to enter spaces at all, this looks weird
- other MUA may have the possibility to filter the password before decrypting.
- only if both does not work on the destination device, a hint is usefull - but then this should be done on the destination device. Should we recommend this hint to the user?

doc/level1.rst
```diff
@@ -161,42 +161,40 @@ NOT be included in the cleartext of the Autocrypt Setup Message, or
...
-two characters of the setup code in the ``Passphrase-Begin`` header so that the receiving MUA can display
-a message as "Please type in the Setup code starting with 12..."::
+If those digits are included in the headers like this, they may also
+be used in the descriptive text that is part of the Setup Message, to
+distinguish different messages.
 
-    Passphrase-Begin: 12
-    
-Non-Ascii characters in the ``Passphrase-Begin`` header MUST be encoded using the 
-encoded-word syntax (RFC 2047).  This becomes important when free UTF-8 strings are used and not `numeric9x4`.
```
r10s: From the view of an implementor, I think it is not automatically clear how to decode the string - even if it contains only ASCII numbers at the moment, this may change soon, esp. as we say the setup code uses UTF-8 above, it may happen, `Passphrase-Begin` is also an UTF-8 string.

However, if it is implicitly clear by other standards, I'm fine with leaving it out.

Valodim: It is not implicitly clear, unfortunately. But the thought here was that a generated passphrase would never use characters that aren't ascii printable, and user-provided passwords wouldn't be partly revealed using this header.

r10s: Sure, for now, there is no problem. And I'm fine in leaving the line out. 

My thoughts were the following:

* It may happen for some reason, other `Passphrase-Format`-values are defined that really use UTF-8 - and then there will be a problem with `Passphrase-Begin`-implementations that do no decode-word.

* `Passphrase-Begin`  may also get useful without  `Passphrase-Format`.

* `Passphrase-Format` and `Passphrase-Begin` may also get handy for non-Autocrypt-stuff (I thought this was one of the reasons why there is no Autocrypt in the name and why we decided to put it into the PGP-header and not into the MIME-header)

More general:

* If the whole setup code is UTF-8, parts of it should be, too.

* In practice, there is already a decode-encoded-words parser present, and the implementor has to choose whether to use `begin=header('Passphrase-Begin')` or `begin=decord_words(header('Passphrase-Begin'))`

Valodim: openpgp implementations aren't always tied to e-mail clients. an rfc2047 parser would be a dependency we introduce, while we don't actually have a need for it.

As an alternative suggestion, how about we tie the begin more tightly to the format?

```
Passphrase-Format: numeric9x4; begin=95
```

r10s: Okay, I understand. In fact, requiring an rfc2047-parser _only_ to implement this is overkill.

However, using `Passphrase-Format: numeric9x4; begin=95` also introduces the requirement for additional parsers ...

So maybe it is the best to stop discussion here and just stuck to the existing spec :-) I would be fine with that.

Valodim: Cool :+1:

PR #112, opened 2017-06-15
Correct the old 'alphanumeric' and 'key' identifiers to their new names, update 'Happy path' example .

r10s:
Just two typos, update happy path example.


r10s:
Intentionally, I wanted to create a separate commit for the "(either `mutual` or `nopreference`)" change in level1.rst in Update 'Happy path' example. However, now it's here ...

On topic: I think, we've never mentioned the value-names of the user's prefer-encrypt-setting - but if it's a MUST to add the Autocrypt-Prefer-Encrypt header, we also need a name for the non-mutual setting.

PR #117, opened 2017-06-20
change motto to "encrypted e-mail for humans"

hpk42:
as it has a nicer ring to it and also piggypacks on a "for humans"
meme widely used in the python community to signal convenient APIs,
see e.g. https://pypi.python.org/pypi?%3Aaction=search&term=for+humans+&submit=search
This relates to the aim of Autocrypt to provide "convenient" and
"usable" e-mail encryption.


r10s:
I like the new motto. It works also without knowing that it is widely used in some communities.

dkg:
some of my best friends are humans.  I'm down with this change.

👍

PR #122, opened 2017-07-04
spec: rephrase multi-device section in intro

azul:
With the Setup Message we now somewhat support multi device
configurations - but not sync. Trying to be more precise here.
Also... next is level 2 - not level 1.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -10,8 +10,8 @@ mails while supporting users who want to explicitly turn on
 encryption.
 
 For ease of implementation and deployment, Level 1 does not support
-multi-device configurations.  We intend to :doc:`support the multi-device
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -10,8 +10,8 @@ mails while supporting users who want to explicitly turn on
 encryption.
 
 For ease of implementation and deployment, Level 1 does not support
-multi-device configurations.  We intend to :doc:`support the multi-device
-use case (and other features) as part of Level 1<next-steps>`.  We
+multi-device synchronisation.  We intend to :doc:`support the multi-device
```
hpk42: i am merging this but think that we could say that Level 1 only supports manual multi-device synchronisation where a user can transfer a secret from one device to another ...

Issue #102, opened 2017-06-13
Document why we skip on more than one valid header?


azul:
(from a todo in The Autocrypt Header on line 390 of level1.rst)

dkg:
The goal is simplicity, and it's not clear that there is a motivation for a client to want to send multiple keys of the same type for themselves.

So, we skip on more than one valid header per `<peer,type>` because:

 * conformant peers will send only a single header, so this looks like something is wrong.
 * if we see more than one header, then we have to make a decision about what to do with the extra headers.  store them all?  store the first one?  store the last one?
 * if we choose to store all the headers found per `<peer,type>` tuple, then replacement/combination/counting becomes significantly more complicated from the state management perspective.  What happens when you get message A with headers X,Y and prefer-encrypt=mutual and then message B with headers X,Z and no prefer-encrypt ?
 * if multiple headers are stored per `<peer,type>` tuple, then making an autocrypt recommendation is also more complicated.  Do we recommend multiple keys for encryption?
 * in the future, if there is some sort of "autocrypt address" or "autocrypt number" for peer key confirmation, how will it look if there are multiple keys for a given `<peer,type>` tuple?

What is the use case where a peer might want to send multiple keys of the same type for themselves?

dkg:
I think this is now resolved, since the merge of #123

PR #123, opened 2017-07-04
spec: move handling multiple headers into parse section

azul:
The implementors need to take this into account when implementing the
parsing of an incoming message. It's not really part of the definition
of an autocrypt header itself.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -35,11 +35,15 @@ we opted for readability.
 
 
 Why do you drop all headers if there is more than one?
-------------------------------------------------------
+-------------------------------------------------------------
 
-Because of multi-agent usage we may have to handle an inconsistent stream of
-headers already. Making this an inconsistent stream of multiple keys with
-priorities sounds like a lot of pain.
+We could come up with rules which header to pick. But whatever we do,
+it has to be deterministic, clear and agreed upon by all clients so the
+behaviour is predictable.
+Dropping all headers is the simplest way to avoid an ambiguous state in
+level 1. Once we have more experience from the field we'll know how this
+fails and at that point we'll be in a position to draft more complicated
+rules.
 
 
 What if I want my MUA to announce two different keys?
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -106,6 +106,9 @@ minimal Level 1 MUA will only include these two attributes.  If
 ``own_state.prefer_encrypt`` is set to ``mutual`` then the header MUST
 have a ``prefer-encrypt`` attribute set to ``mutual``.
 
+The client MUST NOT include more than one valid Level 1 ``Autocrypt``
+header.
+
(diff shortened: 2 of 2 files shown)


azul:
Okay... so now the spec says mails MUST NOT include more than one valid autocrypt level 1 header.
And clients SHOULD drop all headers if they do.

Dealing with such a situation (that should not occure) differently than by dropping all headers would not break Autocrypt entirely... thus this is a SHOULD now.

Also answered the corresponding question in the FAQ in a bit more detail including @BjarniRunar s comment on irc.

azul:
fixes #102

doc/faq.rst
```diff
@@ -34,12 +34,15 @@ An ECC key is roughly 500 bytes formatted in Base64 and RSA 2048 key is
...
-
-Because of multi-agent usage we may have to handle an inconsistent stream of
-headers already. Making this an inconsistent stream of multiple keys with
-priorities sounds like a lot of pain.
+Why do you ask to drop all headers if there is more than one?
+-------------------------------------------------------------
+
+Because of multi-agent usage we may have to handle an inconsistent
+stream of headers already. Making this an inconsistent stream of
+multiple headers per mail sounds like a lot of pain. We could come up
```
hpk42: actually, i never could quite follow this reasoning or this wording.  How does "an inconsistent stream of multiple headers" come about in the "there is more than one header" case?  Shouldn't this answer rather be about the difficulty of choosing one of multiple headers?  Let's say i get two valid Autocrypt headers in one mail and i choose which one i am going to use randomly -- what exactly would be the problem?

Valodim: I think this is a Postel's law thing. If we receive two valid autocrypt headers, clearly *something* must be going on. If we are "liberal in what we accept" here, e.g. by choosing one at random, that means clients can get away with breaking the MUST NOT in regard to sending multiple headers that can be treated as equally valid (while we have no defined order of preference). If we do that, we might as well soften that MUST criteria.

BjarniRunar: @hpk42 , the problem is that the overall behaviour of the system becomes unpredictable. That is never a desirable quality. To avoid it being unpredictable, we would need to clearly specify the rules for choosing a header, which adds complexity. Also not a desirable thing.

Making it clear that this is a spec violation and ignoring it completely is simple and in the exact same spirit as our rules about parsing the headers themselves: if mandatory attributes are seen that we don't recognize, we ignore the header because we cannot be confident that we're doing the right thing. The same logic applies here: if we see multiple headers, we cannot be confident that choosing any particular one is the right thing - so we error out and do nothing.

It's clear. Once we have more experience from the field we'll know *how* this fails and at that point we'll be in a position to draft more complicated rules. But at this early stage, we just don't know enough and should strive to keep things simple.

In my opinion, obviously... ;-)

PR #124, opened 2017-07-04
spec: simplify the message processing a bit

azul:
First of all we now catch messages without a header and an
effective message date that is equal to last_seen.

In addition this separates a long combined clause into two.
We can compare the message date to last_seen_autocrypt right
at the beginning because last_seen_autocrypt is always older
or equal to last_seen and so if a message is older than
last_seen_autocrypt we do not need to process it at all.


Valodim:
:+1:

PR #125, opened 2017-07-09
remove inner MIME structure for self-sent message

Valodim:
The decision to have a MIME structure inside the self-sent message encrypted data seems to be a little shaky. I went ahead and created an alternative suggestion as a basis for discussion.

This does *not* include sync for the prefer-encrypt preference. I don't think that's a big problem.


Valodim:
A more important point for this was brought up by @BjarniRunar: The inner MIME structure contains ASCII-armored data, which makes it tempting to just do `gpg --decrypt | gpg --import` instead of parsing MIME. Since the point of MIME is extensibility, this can be considered a bug in the spec.

I also think we don't lose much if we have a trivial format here: if switch to a more complex one at a later point, handling of the trivial format will already be available, and otherwise well, trivial to implement.

hpk42:
I'd like to hear a little more reasoning than just "decision was shaky".  One particular reason for a mime structure that it allows apps to have custom sections for cross-device transfer of  state, like address books, profile pictures.  While this can be re-added later, why remove it now if it's already there and the use case exists in delta-chat?  

**Complexity** arises in my experience from the number of if's one needs to handle and their interrelations.  The current Setup Code format does not have lots of if-s, just a **nested** structure.  Reducing nesting levels is often a good idea, though :) 

FWIW i currently regard the setup message as a potentially longer lasting format as it is kind of a fall back if all else fails (including potential level 2 mechanics).   Changing it later makes implementations more complex because they then would need to handle two cases for formats.

Valodim:
The setup message as it is now requires the user to type 36 digits to import, which makes it fundamentally unsuitable for any kind of "sync" purpose. If we have a complex data format, it is only for consistency with a later format that is also used for sync. Yes, we want to have a sync mechanism at some point. But we hardly thought that through at this point, a decision based on those loosely anticipated requirements indeed feels very "shaky" to me.

Case in point on implementation complexity: At the moment I have a call to OpenKeychain that does some UI and returns the user's encrypted secret key in the trivial openpgp format. Conversely, I can hand that data back to import that secret key. Adding MIME in between there requires either OpenKeychain to know about this particular MIME format and parse it, or K-9 to work with the raw secret key data somewhere in the process (it never does currently) and adding an extra IPC roundtrip. The coupling gets a lot tighter one way or another. Now that's not a huge deal to implement, but it sure is more than nothing.

r10s:
The use case for Delta Chat would be an initial setup (to transfer data eg. to an additional device), not a "daily" sync.

_Today_, I would not use it: The groups names are transferred with every message and there are no profile or group images yet. But this may change soon (if I ask my daughter, this is the most important issue :-). When losing the possibility to add the information to the encrypted payload, maybe I will add them to the unencrypted part. Or add a second encrypted payload. Or add them to the OpenPGP data (this seems to be ugly) (a "real" sync seems not to be needed from the view of Delta Chat, if all devices are set up, all devices get the same messages and should be in sync automatically).

What I do not really understand: Future versions of Autocrypt may transfer the keys in a smarter way, eg. without the need to enter long passcodes. But a sort of key transfer would still be required - so, why not use the same setup file format then?

Regarding the prefer-encrypt preference: IIRC, on the Freiburg meeting, many (?) people agreed, that transferring this settings is needed. In fact, this was the initial reason for the MIME structure IIRC. So, with or without inner MIME, I think we should transfer this setting in some way - otherwise the receivers may get prefer-encrypt=mutual from the one device and a no-preference from there other.

On the other hand, from the view of implementors, I also see that the Setup message seems be be too complicated (although I have more problems with the symmetric encryption than with the MIME part ...).

BjarniRunar:
I am quite strongly of the opinion that MIME is not a nice format to work with. As e-mail developers we know this better than most; we should know better than to encourage its use.

Nearly any other structured data format would be preferable - including XML and JSON. As an extensible container, using ZIP would be my preference, simply because it has widespread support and tooling, both at the cross-platform OS level and in software development libraries. Precendent: OpenDocument files are .zip files, JARs are .zip files, Python wheels are .zip files, etc. etc. None of those projects decided to use MIME as a container. :-)

The idea that "we're already doing MIME so adding more MIME parsing is free" is refuted by Vincent's point above.

When I brought this up on IRC, people seemed to generally agree that more MIME parsing isn't great, we've already identified some plausible traps doing so would lay for the unwary developer. Coming to a consensus on a different container seemed like an uphill battle, but just simplifying and removing the container layer entirely was not controversial, thus this PR.

r10s:
What exactly was the point refuting that "MIME parsing is free if one has already a MIME parser"?

I also do not like MIME, but _if_ we decide that we need a more complex payload, from my point of view, using an existent MIME parser is easier that using ZIP/JSON/XML with an extra parser only for this purpose. However, the reasons for such a change should not be _removing complexity_ ;-) (in fact, I think such a change would make implementing Autocrypt for implementors much more difficult)

So, I would be fine in removing the inner MIME structure if we don't need it in the foreseeable future. But I would keep it, if it is clear that we'll need the complexity in six months. And I would not replace MIME by something else.

Valodim:
With the "mutual" semantics, the prefer-encrypt setting only affects messages where all parties have it enabled, i.e. advanced users, so the flip-flop state isn't as bad as it was before. Transferring the prefer-encrypt preference in this incomplete way is a "nice to have" for me that is worth very little sacrificed simplicity.

Other than for this particular purpose, I don't think we need the MIME structure in the session setup message, ever. We might need it in a different message that we actually use for sync, later on. But I don't think we'll be doing ourselves any favors by introducing an incomplete format now that we'll later extend, compared to starting from scratch.

r10s:
Regarding the prefer-encrypt setting:

I think it is needed to transfer the state at least if the default is `prefer-encrypt=mutual` instead of the recommended `prefer-encrypt=nopreference` (for Delta Chat, I default to encryption and ignore [this SHOULD](https://autocrypt.org/en/latest/level1.html#account-preferences)).

If the user switches encryption off on DeviceA and transfers the key to DeviceB, this may result in unexpected encrypted messages as on DeviceB the encryption defaults to on.

So, independently of the format, we should transfer the prefer-encrypt setting in some way.

Valodim:
Go by an outer autocrypt header, maybe? Or guess by "if user-agent != delta-chat"? I don't think it's particularly problematic to transfer this bit of information outside the encrypted container.

Is the setting even exposed to the user in delta-chat? Behaving differently there based on the import setting but not showing the option otherwise might also be confusing ("why aren't messages between our two delta-chats encrypted, but for other people it just works?!")

r10s:
The prefer-encrypt setting is easily editable for the user, it just defaults to mutual.

Valodim:
Did we reach consensus on this issue? I feel like we did, but it's been a while so I'd prefer if someone else confirmed before we merge this :)

hpk42:
i'd like to hear @pbrunschwig 's opinion in particular but also others', also in light of what is implemented with enigmail already.

pbrunschwig:
The changes look alright to me. But it misses the own perfer-encrypt state.

pbrunschwig:
> The setup message as it is now requires the user to type 36 digits to import, which makes it fundamentally unsuitable for any kind of "sync" purpose. 

I'm not sure that this by itself is an issue. We could use for example a Diffie-Hellman based method to transfer the password, and leave the message format as it is.

Issue #126, opened 2017-07-11
Typically unencrypted multipart/report messages force an Autocrypt-reset


r10s:
[multipart/report](https://tools.ietf.org/html/rfc6522) messages seems not to be encrypted in many cases.

This may become a problem when these messages are send eg. for [MDNs](https://tools.ietf.org/html/rfc8098) - if you request a MDN for a message and the receiver sends it without an Autocrypt header we get a peerstate "reset" for the recipient.

The easiest solution may be to require encryption and Autocrypt headers for all MDNs, however, I fear, this is not always possible (eg. in Thunderbird it seems as if the MDN-send-button has nothing to do with Enigmail).

My proposal would be to say that MUAs implementing Autocrypt SHOULD ignore missing Autocrypt headers for multipart/report messages. And MUAs sending multipart/report messages MAY encrypt them.

Valodim:
sounds good to me :+1:

Valodim:
I'd leave "the MUA MAY also encrypt this type of message" out. I thought where to put this detail, but couldn't find a good place and it really feels kind of obvious at this point.

r10s:
Yes, I noticed the missing "MAY encrypt" and I agree in your point that it is obvious that MUAs may encrypt reports (which, however, seems only to be standard-conform since 2012 with RFC 6522; before the mutipart/report must be the outermost part)

PR #127, opened 2017-07-11
add paragraph about multipart/report

Valodim:
see #126

I wanted to avoid making the regular parsing more complicated for a fairly optional SHOULD, so I put it into an extra paragraph. It's not perfectly positioned, but I couldn't find a better place either.


hpk42:
wording could be improved but that's no reason to block this improvement/clarification.

PR #130, opened 2017-08-07
remove these brackets because they are required fields

karissa:
The brackets make it seem like `prefer-encrypt` and `type` are optional -- are they? The documentation makes the impression that they are critical and required.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -120,7 +120,7 @@ The ``Autocrypt`` Header
 
 The ``Autocrypt`` header has the following format::
 
-    Autocrypt: addr=a@b.example.org; [type=1;] [prefer-encrypt=mutual;] keydata=BASE64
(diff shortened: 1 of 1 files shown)


Valodim:
all parameters that don't start with an underscore are "critical", but not necessarily mandatory:

"critical" refers to the property that a parameter must be parsed if present. This is useful to introduce parameters at a later point that change semantics in a way that should trigger a failure case rather than be ignored.

Both `type` and `prefer-encrypt` parameters have default values (`1` and `no-preference`, respectively), so they are indeed optional. I think this becomes slightly clearer for `type` in #132, too.

I think the intent has been clarified here, so I'll be closing this PR. If someone thinks the phrasing could be better for any of these points in the spec, please send a clarification PR that we can discuss (as always: concrete suggestions are easier to discuss than abstract ideas of what could be read into the spec and what could not)

PR #133, opened 2017-09-21
Updated list of gatherings, added panoramix

moba:
With Vincent and me in Panoramix, it might make sense to add it? I do see quite some overlap and shared conversations that matter for the messaging space as a whole.

Events: I added the pretty spontaneous get-together that is happening now, the Berlin idea, 34c3, and moved RWC so it is properly ordered.


Valodim:
lgtm :+1:

PR #142, opened 2017-11-03
fix #141 : be clear on what attributes are mandatory or optional

### LINKED ISSUE #141: Autocrypt header: not clear whether type=1 is an optional or a mandatory attribute
When reading https://github.com/autocrypt/autocrypt/blob/master/doc/level1.rst#L123, ``[`` suggests ``[type=1]`` is an optional attribute. However reading https://github.com/autocrypt/autocrypt/blob/master/doc/level1.rst#L131 and below, gives the impression that this attribute is mandatory.

hpk42:
Go through the Autocrypt header wording and clarify mandatory/optional things.

also fix some grammar, simplify sentences.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -131,33 +131,32 @@ The ``Autocrypt`` header has the following format::
 
     Autocrypt: [type=1;] addr=a@b.example.org; [prefer-encrypt=mutual;] keydata=BASE64
 
-The ``type`` parameter defines the meaning of the rest of the header.
-For now the only defined type is ``1``, which implies the meaning as
-described here. This is also the default, if the header is absent.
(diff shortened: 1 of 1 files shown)


azul:
looks good to me.

dkg:
I've rebased this branch so that it has the possibility of being merged.

while i agree that clarification of "optional" vs. "required" is useful, this patch also makes a substantive change by requiring that the type header moves to the front of the list.  i'd prefer to merge just the "optional" vs. "required" changes, while treating the "move to front" in a separate pull request.

dkg:
I've just opened #149, which is this pull request without the required ordering.  I prefer that we merge #149 instead of this one.

PR #147, opened 2017-11-04
Explicitly suggest RSA 3072-bit key generation.

dkg:


doc/level1.rst
```diff
@@ -760,8 +762,8 @@ Level 1 MUAs maintain an internal structure ``own_state`` for each
 account on which Autocrypt is enabled. ``own_state`` has the following
 members:
 
- * ``secret_key`` -- the RSA 2048-bit secret key used for this
-   account (see "Secret Key Generation and storage" above).
+ * ``secret_key`` -- the 2048- or 3072-bit RSA secret keys used for
```
azul: why is this a plural 'keys' now?

doc/ecosystem-dangers.rst
```diff
@@ -24,7 +24,7 @@ communication.
 Message Deliverability
 ----------------------
 
-:mailheader:`Autocrypt` headers that use RSA 2048 are large enough
+:mailheader:`Autocrypt` headers that use RSA 3072 are large enough
```
hpk42: sidenote: there is no mailheader directive -- just use "``mailheader``"

r10s:
Just tested it, at a first glance, it also works with NetPGP.

And, in general I also feel some more bits are a good idea :)

PR #149, opened 2017-11-04
fix issue141 : be clear on what attributes are mandatory or optional

dkg:
updated by dkg to avoid adding the requirement for "type"

This supercedes #142 -- it resolves the "mandatory" v. "optional" question without introducing an explicit ordering requirement for the type attribute.


dkg:
I prefer this solution over the proposal in PR #142

PR #150, opened 2017-11-04
remove "type" attribute

dkg:
After discussion with @Valodim, i'm convinced that specifying the
type= header in the level 1 spec doesn't gain us anything.  and
removing "type" simplifies the spec.

One thing that "type" gave us was a plan for how to move to a
different key format.  However, we can use that same plan by taking
advantage of a critical attribute, and defining it to declare a new
way to interpret keydata.

Another thing that type gave us (see
a2f31b4ca4ff3cd3e4fc00b7a7de0b405b6afe2c) was a way to establish
forward compatibility if we wanted to introduce a way to include
semicolons in attribute values (e.g. introducing quoting conventions).
I think trying to introduce that level of complexity in headers is a
source of risk.  And we can achieve the same flexibility by declaring
a new header entirely (e.g. Autocrypt2:).

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -129,13 +129,7 @@ The ``Autocrypt`` Header
 
 The ``Autocrypt`` header has the following format::
 
-    Autocrypt: [type=1;] addr=a@b.example.org; [prefer-encrypt=mutual;] keydata=BASE64
-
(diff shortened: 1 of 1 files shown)


Valodim:
lgtm :+1:

hpk42:
i am ok with it, the pr currently has conflicts though.

PR #152, opened 2017-11-04
add high level overview section

Valodim:
I spent some time wording a technical overview  with @dkg, that gives the reader a high level overview of our approach and core ideas of the spec. I don't consider this finished, it's missing a part about how we manage secret keys and possibly references to the chapters that deal with the introduced concepts.

Feedback welcome

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -61,10 +61,10 @@ The core goal of Autocrypt is to fully automate the management of both
 secret and public keys, so that users can encrypt mail without
 specialized knowledge.
 
-An Autocrypt header is introduced that transfers information about a
-sender's public keys in all their emails. Autocrypt provides a simple
-set of rules to track this information per communication peer. This
-provides unambiguous information during composition on whether
+This spec introduces an Autocrypt header that transfers information
+about a sender's public keys in all their emails. Autocrypt provides a
+simple set of rules to track this information per communication peer.
+This provides unambiguous information during composition on whether
 encryption is a) possible and b) recommended for a given set of
 recipients. The design relies on in-band communication for key
 discovery, thereby avoiding a dependency on external infrastructure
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -54,6 +54,34 @@ will not be able to enable Autocrypt on it.
...
+
+Autocrypt facilitates opportunistic key distribution, but recognizes
+that aggressively opportunistic encryption can be disruptive to
+established email workflows. For this reason, emails are only
+encrypted if either:
+
+1) The sender specifically requests encryption during composition
+2) When replying to an encrypted message
+3) If all participants have explicitly expressed a preference to
+   always encrypt.
```
hpk42: s/always/mutually maybe?

doc/level1.rst
```diff
@@ -54,6 +54,34 @@ will not be able to enable Autocrypt on it.
...
 
+Approach and High Level Overview
+--------------------------------
+
+The core goal of Autocrypt is to fully automate the management of both
+secret and public keys, so that users can encrypt mail without
+specialized knowledge.
+
+An Autocrypt header is introduced that transfers information about a
+sender's public keys in all their emails. Autocrypt provides a simple
```
hpk42: i prefer an "active", not passive sentence, something like "Along each sent e-mail, the Autocrypt header transfers public keys and encryption preferences"  (or some other variant)

Valodim: agreed & changed

doc/level1.rst
dkg: I'm not sure we want to use the term "mutual" in the sentence here, because i think it can be subtly confusing.  The preference is expressed unilaterally by each peer, in the hopes that they can encrypt.  the preferences are mutual in the sense that they take effect only when they are expressed by all participants (as point 3 says here), but they are not negotiated in a coordinated fashion (that is, they aren't negotiated pairwise, which might be what people think the word "mutual" means).

So i think the text as it stands is better without "mutual", though i welcome more wordsmithing PRs to improve it further.

PR #154, opened 2017-11-04
Add appendix example and how it could be referenced

juga0:
An example on how we could have an appendix with pseudocode or grammar and reference for the "human text" spec.
Style needs to be fixed.


juga0:
Created branch in this repo and new PR https://github.com/autocrypt/autocrypt/pull/159

PR #156, opened 2017-11-04
starting to tweak the flow after the reordering

azul:
We got stuck with non flow changes though. So this is just the first section...

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -20,7 +20,7 @@ real-life experiences as soon as possible.
 
 Throughout this document, we refer to a mail app or Mail User Agent (MUA)
 as though it was only capable of controlling a single e-mail account
-(see `multiaccounts`_ for more detail).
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -1,26 +1,26 @@
...
+Autocrypt-capable at Level 1.
+
+The design of Level 1 is driven by usability concerns and by the
+realities of incremental deployment. A user may use both
+Autocrypt-enabled mail apps and traditional plain ones and we'd like to
+avoid annoyances like unexpected unreadable mails while supporting users
+who want to explicitly turn on encryption.
+
+For ease of implementation and deployment, Level 1 only contemplates the
+use of Autocrypt on a single device.  We intend to :doc:`support
```
hpk42: it's not "only contemplates" because we also have setup-message key sharing but "Level 1 focus on single-device use" or something.

PR #157, opened 2017-11-04
close #90 (document justification for prefer-encrypt=mutual

### LINKED ISSUE #90: motivate prefer-encrypt=mutual
prefer-encrypt=mutual seems to be weaker than prefer-encrypt=yes.

why did we switch to it?

 * the problem cases arise with situations where peers have different preferences
 * moving the initial encrypted message from one peer to the other doesn't actually resolve the preference mismatch, it just results in surprisingly unreadable mail one hop later in the conversation.
 * surprisingly unreadable mail is likely to create pushback from the party with the weaker preferences, possibly in the form of "please don't send me encrypted mail" messages, which is a strong social pressure against encrypted messages.

dkg:


Valodim:
sounds good

PR #158, opened 2017-11-04
Add table with the autocrypt client implementations status

juga0:
@hpk42's table transcription. Missing some rows at the bottom + mailpile status


dkg:
This table is pretty great.  I think it will help encourage implementers to get things done, and might serve as a step-by-step guide to implementation.  It also sort of implicitly outlines a test suite.

However, it introduces a timeliness question -- can we keep this from going out of date?  how does the reader know whether it is currently accurate or not?

Should we be providing links to the web pages of the specific clients?  can we include a logo for each client? 

Should we be providing versioning information (about when these things are available)?

Should we provide more detailed information about what the rows mean?

Also, we might want to encourage new implementations to submit patches to advertise their implementations.  If we want clean pull requests, perhaps the table should be "flopped" so that a changeset adding a new implementation modifies one or two lines, rather than editing every line.

doc/_templates/globaltoc.html
```diff
@@ -5,6 +5,7 @@
   <li><a href="{{ pathto('index') }}">Introduction</a></li>
   <li><a href="{{ pathto('features') }}">Features</a></li>
   <li><a href="{{ pathto('level1') }}">Level 1 spec</a></li>
+  <li><a href="{{ pathto('implementations') }}">Client implementations status</a></li>
```
dkg: I think i'd just go with "client implementation", not "client implementations".

juga0: Last commit fix some of your comments. Leyend still needs to be created, versions needs to be added, and we don't have still a way to keep it up to date

juga0: @Valodim @pbrunschwig @BjarniRunar could you add your MUA logos in the table?.
Maybe we need to create a new column or just replace the name for an icon.
Thanks

doc/contents.rst
```diff
@@ -26,6 +26,10 @@ and contributors, MUA developers and privacy enthusiasts.
      a first collection of current autocrypt developments (Jan 2017) with
      various mail programs and some development resources.
 
+:doc:`implementations`
+    table with Autocrypt-capable MUAs level 1 implementation status
+    (Nov 2017)
```
dkg: Having the date here gives an opportunity for it to get out of sync with the date next to the table itself.  I'd drop the date from the contents.

doc/implementations.rst
```diff
@@ -0,0 +1,28 @@
+Autocrypt-capable MUAs level 1 implementation status
+=====================================================
+
+Last updated: ``2017/11/03``
```
dkg: I recommend using ISO 8601 here: ``2017-11-03``

doc/implementations.rst
```diff
@@ -0,0 +1,28 @@
...
+================= ======== ======== ======== ======== ======== ======== ======== ======== ========
+ `delta-chat`_    y        y        y        y        y        y        started  y        y
+ `k9`_            y        y [1]_   y        y        y        y        branch   branch   y
+ `enigmail`_      y        y        y        y [2]_   n        y        y        y        n
+ `py-autocrypt`_  y        y        y        y        n        n        n        n        y
+ `notmuch`_       y        n        n        n        n        n        n        n        y
+ `mailpile`_
+================= ======== ======== ======== ======== ======== ======== ======== ======== ========
+
+Leyend: TBD
```
dkg: s/Leyend/Legend/

If TBD means the same thing as "TODO", then we should use "TODO", so that it's easier to find when searching for other TODO items.

doc/_templates/globaltoc.html
dkg: I'd still recommend s/Client implmentations/Client implementation/ -- that doesn't seem to be resolved yet.

juga0:
Rebased, comments should be fixed now.

PR #159, opened 2017-11-04
Features/appendix

juga0:


juga0:
Appendix is now at the end of level1.rst file: https://github.com/autocrypt/autocrypt/blob/master/doc/level1.rst#id122

PR #160, opened 2017-11-04
Improve text flow after section reorganization.

dkg:


doc/level1.rst
```diff
@@ -87,6 +87,33 @@ Autocrypt Internal State
...
+* ``last_seen``: UTC timestamp of the most recent effective date of
+  all processed messages for this peer.
+* ``last_seen_autocrypt``: UTC timestamp of the most recent effective
+  date of all processed messages for this peer that contained a valid
+  Autocrypt header.
+* ``public_key``: the public key of the recipient
+* ``state``: a quad-state: ``nopreference``, ``mutual``, ``reset``, or
+  ``gossip``.
+
+How this information is managed is covered in :ref:`peer-state-management`.
```
Valodim: "managed and used during message composition"?

doc/level1.rst
```diff
@@ -512,7 +534,7 @@ with more than one recipient. These headers MUST be placed in the root
 MIME part of the encrypted message payload. The encrypted payload in
 this case contains one Autocrypt-Gossip header for each recipient,
 which MUST include ``addr`` and ``keydata`` attributes with the relevant
-data from the recipient's Autocrypt peer state.
+data from the recipient's Autocrypt :ref:`peer state <peer-state>`.
```
azul: The recipient's AC peer state sounds like it would be taken from the recipient.
Alternative:
'data from the senders Autocrypt peer state about the recipient'

PR #173, opened 2017-11-04
add a note about use of critical attributes for future incompatible c…

### LINKED ISSUE #163: Suggest how we imagine critical attributes will be used
@Valodim suggests something like this above "OpenPGP Based key data":

Future Autocrypt specs can use critical attributes to force other level 1 clients to reject the header.

Valodim:
…hanges (closes #163)


azul:
@dkg, tried to address your review. Do my changes fix the issue?

PR #176, opened 2017-11-05
clean up a few todos and explain why we need to keep peer state

dkg:


doc/level1.rst
```diff
@@ -249,40 +249,55 @@ Internal state storage
...
-   Explain why we keep peer state.  because we want to handle weird cases like the following...
-
-If a remote peer disables Autocrypt or drops back to using a
-non-Autocrypt MUA only we must be able to disable sending encrypted
-mails to this peer automatically.  MUAs capable of Autocrypt Level 1
-therefore MUST store state about the capabilities of their remote
-peers.
+Autocrypt clients keep state about their peers, to be able to handle
+several nuanced situations that have caused trouble/annoyance in the
+past.  This state is updated even when the peer does *not* send mail
```
Valodim: Well, it is updated when they send mail with no Autocrypt header. It is not updated when they don't send mail at all.

doc/level1.rst
```diff
@@ -249,40 +249,55 @@ Internal state storage
...
+``peer_state[A].state`` semantics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+The ``state`` variable of a particular peer's ``peer_state`` data is
+selected from a set range of values:
+
+  - ``nopreference`` means the peer has not opted into mutual
+    encryption.  The client may or may not know a key for such a peer.
+  - ``mutual`` means we know a key for the peer, and the peer has also
+    opted into mutual encryption.
```
Valodim: "mutual encryption" by itself does not quite cover this. I think this is more clearly phrased somewhere else, something like "expressed agreement to encrypt by default if all parties consent".

dkg:
I think all concerns raised by @Valodim have been resolved.

PR #180, opened 2017-11-05
Complete example Autocrypt Setup Message in appendix

juga0:


doc/level1.rst
```diff
@@ -934,46 +934,243 @@ Example Setup Message
...
-   Content-Type: application/autocrypt-key-backup
-   Content-Disposition: attachment
+    To set up your new device for Autocrypt, please follow the instuctions that should be presented by your new device.
 
-   Possibly a descriptive text, informing users about this file's
-   contents.
-   -----BEGIN PGP MESSAGE-----
-   Passphrase-Format: numeric9x4
-   Passphrase-Begin: 31
+    You can keep this message and use it as a backup for your secret key. If you want to do this, you should write down the password and store it securely.
```
dkg: Why are we using "password" here?  shouldn't this be "setup code" ?

doc/level1.rst
```diff
@@ -934,46 +934,243 @@ Example Setup Message
...
+    --Y6fyGi9SoGeH8WwRaEdC6bbBcYOedDzrQ
+    Content-Type: application/autocrypt-setup
+    Content-Disposition: attachment; filename="autocrypt-key-backup.html"
```
dkg: why is filename="autocrypt-key-backup.html" ?  shouldn't this be "autocrypt-setup-message.html" for consistency in interface?

Valodim: I would omit the html trick from this example. It's not obvious how this works or why it is useful, and in the spec we only mention it in reference to a thought in an external document.

juga0:
Last commit fixes @dkg comments

PR #181, opened 2017-11-05
Add example Setup Message User Interaction in appendix

juga0:


doc/level1.rst
```diff
@@ -980,4 +980,12 @@ Example Setup Message User Interaction
 
 .. todo::
 
-   TODO
+   Is fine just with this?
```
dkg: This line doesn't seem like it belongs in the commit.

doc/level1.rst
```diff
@@ -980,4 +980,12 @@ Example Setup Message User Interaction
...
+
+.. this code decrypts the example Autocrypt Setup Message
+
+::
+
+    This is your Setup Code to decrypt the Autocrypt Setup Message:
+
+    1742-0185-6197-1303-7016-8412-3581-4441-0597
```
dkg: I think we're recommending to present the setup code to the user in the grid format, for easier transcription and copying.

doc/level1.rst
```diff
@@ -980,4 +980,12 @@ Example Setup Message User Interaction
...
 .. todo::
 
-   TODO
+   Is fine just with this?
+
+.. this code decrypts the example Autocrypt Setup Message
+
+::
+
+    This is your Setup Code to decrypt the Autocrypt Setup Message:
```
dkg: should this be "the Setup Code to decrypt your Autocrypt" (swapping "the" and "your")?  this makes it sound cleaner to me.

PR #182, opened 2017-11-05
added section for Onboarding (now called "Helping Users getting Started")

pbrunschwig:


doc/level1.rst
```diff
@@ -878,12 +878,43 @@ Level 1 MUAs MUST allow the user to disable Autocrypt completely for
...
 Autocrypt disabled by default.
 
-Onboarding
-++++++++++
 
-.. todo::
+Helping Users getting Started
++++++++++++++++++++++++++++++
+
+The following paragraph provides recommendations for MUA
```
dkg: paragraph?

doc/level1.rst
```diff
@@ -878,12 +878,43 @@ Level 1 MUAs MUST allow the user to disable Autocrypt completely for
...
+
+2. If a sent message with an Autocrypt header was found:
+
+   Provide guidance for creating an Autocrypt Setup Message
+   on the MUA that created the message.
+
+3. If there is evidence of actively used OpenPGP software
+   (for example if a secret key is available, some
+   specific software is installed, etc.) or if encrypted
+   mails are found:
```
dkg: Is point 3 here intended to mean "non-autocrypt" only?  or would encrypted messages from autocrypt software also trigger this point?

doc/level1.rst
```diff
@@ -878,12 +878,43 @@ Level 1 MUAs MUST allow the user to disable Autocrypt completely for
...
+
+The MUA SHOULD scan the mailbox for sent mails (wherever
+they might be) that show evidence of OpenPGP/Autocrypt
+usage.
+
+
+1. If an Autocrypt Setup Message was found:
+
+   Start a setup process suggesting the user to import the
+   settings.
```
dkg: s/settings/setup message/, right?

doc/level1.rst
```diff
@@ -878,12 +878,43 @@ Level 1 MUAs MUST allow the user to disable Autocrypt completely for
...
+
+The following paragraph provides recommendations for MUA
+implementations to help users start Autocrypt after an
+account was set up.
+
+The MUA SHOULD scan the mailbox for sent mails (wherever
+they might be) that show evidence of OpenPGP/Autocrypt
+usage.
+
+
```
dkg: Is the series below a "first-match-wins" situation?  or should the MUA try them all in parallel?

If the search takes a long time, do we need to exhaust the search first, and then decide what to do based on what we find, or should we take these actions as we find stuff?

doc/level1.rst
pbrunschwig: The numbered list is intended as sequence to follow for all found message. You need to keep a state for it.

PR #184, opened 2017-11-05
Features/unify terms

juga0:


doc/transfer-format.rst
```diff
@@ -41,7 +41,7 @@ Setup Code ``D9VN-RD7A-7T3B-BWG7-LEPY-3NYT``.
   <html><body><h1>Autocrypt Key for <tt>example.org</tt></h1><p>
 
   This is an Autocrypt Key file. When imported into an Autocrypt capable E-Mail
-  client with its backup code, it can be used to read encrypted E-Mail sent to
```
dkg: "E-mail MUA"?

doc/transfer-format.rst
```diff
@@ -107,4 +107,4 @@ Setup Code ``D9VN-RD7A-7T3B-BWG7-LEPY-3NYT``.
   xUbOVydoyTLaP9bQScqv4WBqVcw5k8v4i+I=
   =k6Rc
   -----END PGP MESSAGE-----
-  </pre><hr><i><a href="http://example.org/">Generated by Autocrypt-capable Client</a>.</i></body></html>
+  </pre><hr><i><a href="http://example.org/">Generated by Autocrypt-capable MUA</a>.</i></body></html>
```
dkg: This looks like user-facing stuff -- we don't want to include "MUA" here, do we?  This is a fake ad/promotion for the "example" MUA.  We should make the ad look more plausible than this.

PR #186, opened 2017-11-05
drop doc/transfer-format.rst

dkg:
the separated transfer format documentation no longer matches the
spec.

the spec is also significantly more detailed than it used to be, so i
don't think we need to keep the separate document around any more.


Valodim:
This format was never intended to be part of the spec, but an external recommendation about how the format could be augmented as html for improved "external" usability, i.e. to be more self-descriptive when users save it as a file. We decided against having those details in the spec, but that they are still useful to have around, and hence moved them to this external document.

Did this change? Was this background taken into account in the consideration to delete it?

PR #195, opened 2017-11-05
don't put html in the setup message example

Valodim:
I feel like having html in the example is distracting. In particular,
it's not obvious why the content-disposition would have
filename="something.html", and we never mention it in the spec.

See also https://github.com/autocrypt/autocrypt/pull/186

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -1023,17 +1023,13 @@ Example Setup Message
     and store it securely.
     --Y6fyGi9SoGeH8WwRaEdC6bbBcYOedDzrQ
     Content-Type: application/autocrypt-setup
-    Content-Disposition: attachment; filename="autocrypt-setup-message.html"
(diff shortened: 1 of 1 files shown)


dkg:
This change is small enough that i don't think it saves any significant confusion.  If the concern is that there's no suggestion for HTML in the definition of the Autocrypt Setup Message, i'd prefer adding a single sentence to the definition of the Autocrypt Setup Message.

Valodim:
¯\_(ツ)_/¯

juga0:
@dkg you mean keeping the .html and adding a sentence in the specs?, isn't already there? (https://github.com/autocrypt/autocrypt/blob/6e56bf4fdf99ef457e178697c9320b29d782ad3a/doc/level1.rst)

juga0:
oh, @Valodim, you just removed the .html...

PR #198, opened 2017-11-05
Cleartext replies

Valodim:
I slightly changed the section about cleartext replies, removing the inline text suggestion and mentioning key gossip. I also moved it below the "Mail Encryption" headline.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -474,31 +474,6 @@ should not disable encryption without communicating this to the user.
 A graceful way to handle this situation is to save the enabled state,
 and only prompt the user about the issue when they send the mail.
 
-Cleartext replies to encrypted mail
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-In the common use case, a reply to an encrypted message will also be
-encrypted. Due to Autocrypt's opportunistic approach to key discovery,
-however, it is possible that the ``peers`` entry for one of the
-recipients may be missing, or that it is present, but the ``keydata``
-is missing, which means the reply can only be sent in the clear.
-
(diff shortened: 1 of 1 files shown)


dkg:
I'm willing to accept with this change, though i would prefer to have an example of the "note" that should be "left".

doc/level1.rst
```diff
@@ -527,6 +501,26 @@ partially composed messages to their e-mail provider (e.g., in the
...
+
+Cleartext replies to encrypted mail
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the common use case, a reply to an encrypted message will also be
+encrypted. Due to Autocrypt's opportunistic approach to key discovery,
+however, it is possible that keys for all recipients are not
+available, and a reply can only be sent in the clear.
+
+A MUA that is relatively more oriented towards privacy aware users MAY
```
dkg: maybe "privacy-aware users" ?

doc/level1.rst
```diff
@@ -527,6 +501,26 @@ partially composed messages to their e-mail provider (e.g., in the
...
+prepare the cleartext reply without including any of the quoted and
+attributed text from the previous message, to avoid leaking the
+cleartext from the original encrypted message. In this case, a note
+should be left that the content was redacted, to avoid confusing the
+user.
```
dkg: Confusing the user who is composing the message is one thing.  The note might itself also be oriented toward the recipient of the message, right?  reducing their confusion is also a good rationale.

Valodim: This depends on how the MUA can display this information, I imagine that many will have no other option than placing it inline.

Do you have a concrete change in mind here?

PR #200, opened 2017-11-05
Minor (hopefully) textual improvements.

nwalfield:


doc/level1.rst
```diff
@@ -384,25 +384,24 @@ overall. Additionally update the state as follows:
...
 That said, all Autocrypt-capable MUAs should be able to calculate
-the same Autocrypt recommendation due to their internal state.
+the same Autocrypt recommendation.
 
-The Autocrypt recommendation depends on the list of recipient
-addresses for the message being composed.  When the user edits the
-list of recipients, the recommendation may change.
+The Autocrypt recommendation depends on the recipient
+addresses of the draft message.  When the user changes the
+recipients, the Autocrypt recommendation may change.
```
Valodim: list of?

doc/level1.rst
```diff
@@ -325,18 +323,18 @@ Updating Autocrypt Peer State
...
 
-  - The content-type is ``multipart/report``. It can be assumed to be
-    auto-generated. This in particular avoids triggering a ``reset``
+  - The content-type is ``multipart/report``. In this case, it can be assumed
+    the message was auto-generated. This avoids triggering a ``reset``
     state from received Message Disposition Notifications (:rfc:`3798`).
 
   - There is more than one address in the ``From`` header.
 
   - The MUA believes the message to be spam. If the user marks the
```
Valodim: is the second sentence here even useful? seems to me we don't lose much by dropping it

doc/level1.rst
```diff
@@ -128,31 +125,32 @@ Accounts controlled by the MUA
...
+How this information is managed and used is discussed in :ref:`own-state-management`.
 
 .. _peer-state-management:
 
 Peer State Management
 ---------------------
 
-Autocrypt MUAs update state about their communications peers based
-on information gathered from received e-mail headers.
+An Autocrypt MUA updates each communication peer's state using the
```
Valodim: "the state they keep about each communication peer", maybe? the "peer's state" sounds like changing something about the key?

doc/level1.rst
```diff
@@ -128,31 +125,32 @@ Accounts controlled by the MUA
...
+ * ``secret_key``: The RSA secret key material used for
+   the account (see :ref:`secretkeys`).
+ * ``public_key``: The OpenPGP transferable public key (:rfc:`OpenPGP
+   "Transferable Public Key"<4880#section-11.1>`) derived
    from the secret key.
- * ``prefer_encrypt`` -- the user's own
-   preferences on the account, either ``mutual`` or ``nopreference``.
-   This SHOULD be set to ``nopreference`` by default.
+ * ``prefer_encrypt``: The user's encryption 
+   preferences for this account.  This is either ``mutual`` or ``nopreference``.
```
Valodim: "preference" singular?

doc/level1.rst
```diff
@@ -67,24 +67,21 @@ encryption is possible, and:
...
-has credentials and capabilities to perform these network services:
+Autocrypt tries to impose minimal requirements on :term:`MUA` and
+e-mail service interactions.  Specifically, an Autocrypt-capable MUA
+needs to be able to:
 
-- The ability to send e-mail (e.g. via SMTP or Submission) where the
-  MUA can control the entire message being sent, including both
-  message headers and message body.
+- Control the contents of outgoing e-mail including the ability to set
+  custom e-mail headers;
```
Valodim: this drops the requirement to send messages on its own, which is required for the setup message. I still like the change, maybe we can just move sending messages to a different point

doc/level1.rst
```diff
@@ -67,24 +67,21 @@ encryption is possible, and:
...
 
-- The ability to receive e-mail where the MUA gets access to the
-  entire message being received, including both message headers and
-  message body.
+- Read the whole, raw e-mail including message headers; and,
```
Valodim: "the" e-mail?

doc/level1.rst
nwalfield: ack.

doc/level1.rst
nwalfield: ack

doc/level1.rst
nwalfield: ack

doc/level1.rst
nwalfield: ack

doc/level1.rst
nwalfield: nack.  This changes the substance, I think.

doc/level1.rst
nwalfield: nack. list of is unnecessary.

PR #202, opened 2017-11-05
Rename own_state and peer_state to account_data and peer_data

Valodim:
As discussed previously. I'm not sure if this is a super great idea yet, but creating this PR as a basis for discussion.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -96,19 +96,19 @@ Autocrypt Internal State
 An Autocrypt MUA needs to associate information with the accounts it
 controls and the peers it communicates with.
 
-.. _peer-state:
+.. _peer-data:
 
 Communication Peers
 ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


Valodim:
I'm not sure about "peer data", but renaming the "own" stuff to "account" stuff feels like an improvement to me

dkg:
I think this reorganization is great, but it doesn't go far enough.

What if we made `account_data` itself a table indexed by account e-mail address, in the same way that `peer_data` is a table indexed by peer e-mail address?

At that point, we can remove some of the caveats like "as though it were only capable of controlling a single e-mail account" and everything becomes neatly more parallel.

dkg:
@azul points out that we could go even simpler, and just have two tables for the MUA, both indexed by e-mail address:

 * `accounts`
 * `peers`

then we don't need the words `_data` or `_state` at all.

azul:
superseeded by #204

PR #204, opened 2017-11-05
replace own-state -> accounts, peer-state -> peers

azul:
superseedes #202

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -46,7 +46,7 @@ information, the sender's :term:`public key`.  Transferring public
 keys in-band means that key discovery in Autocrypt does not require
 external infrastructure like OpenPGP keyservers or x509 PKI.
 
-Autocrypt provides a :ref:`set of rules <update-peer-state>` that
+Autocrypt provides a :ref:`set of rules <update-peers>` that
 tracks this information for each communication peer.  Autocrypt uses
 this information to determine whether encryption is possible and makes
 a :ref:`recommendation <recommendation>` about whether encryption
@@ -96,18 +96,18 @@ Autocrypt Internal State
 An Autocrypt MUA needs to associate information with the accounts it
 controls and the peers it communicates with.
 
-.. _peer-state:
+.. _peers:
 
 Communication Peers
 ~~~~~~~~~~~~~~~~~~~
 
 An e-mail address is a communication peer.  Autocrypt associates state
 with each peer.  Conceptually, we represent this state as a table named
-``peer_state`` indexed by the peer's :doc:`canonicalized
+``peers`` indexed by the peer's :doc:`canonicalized
 e-mail address <address-canonicalization>`.
 
 For each e-mail address ``A``, an MUA MUST associate the following
-attributes with ``peer_state[A]``:
+attributes with ``peers[A]``:
 
 * ``last_seen``: The UTC timestamp of the most recent effective date
   (:ref:`definition <effective_date>`) of all messages that the MUA has
@@ -121,14 +121,15 @@ attributes with ``peer_state[A]``:
 
 How this information is managed and used is discussed in :ref:`peer-state-management`.
 
-.. _own-state:
+.. _accounts:
 
 Accounts controlled by the MUA
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A Level 1 MUA maintains an internal structure ``own_state`` for each
-account on which Autocrypt is enabled. ``own_state`` has the following
-attributes:
+A Level 1 MUA maintains an internal structure ``accounts`` indexed by
+the  the account's :doc:`canonicalized e-mail address
+<address-canonicalization>` (`A`).  For each account on which Autocrypt is
+enabled, ``accounts[A]`` has the following attributes:
 
  * ``secret_key``: The RSA secret key material used for
    the account (see :ref:`secretkeys`).
@@ -140,11 +141,11 @@ attributes:
    This SHOULD default to ``nopreference``.
 
 If Autocrypt is enabled for a given account, the MUA SHOULD allow the
-user to switch the setting for ``own_state.prefer_encrypt``.  This
+user to switch the setting for ``accounts[A].prefer_encrypt``.  This
 choice might be hidden in something like a "preferences pane".  See
 :ref:`preference-ui` for a specific example of how this could look.
 
-How this information is managed and used is discussed in :ref:`own-state-management`.
+How this information is managed and used is discussed in :ref:`accounts-management`.
 
 .. _peer-state-management:
 
@@ -233,16 +234,16 @@ Header injection in outbound mail
 
 During message composition, if the ``From:`` header of the
 outgoing e-mail (the ``sender address``) matches an address for which the Autocrypt-capable MUA
-has secret key material (``own_state.secret_key``), it
+has secret key material (``accounts[A].secret_key``), it
 SHOULD include an Autocrypt header. This header MUST contain the
-corresponding public key material (``own_state.public_key``) as the ``keydata``
+corresponding public key material (``accounts[A].public_key``) as the ``keydata``
 attribute, and the ``sender address`` as the ``addr`` attribute.  The most
 minimal Level 1 compliant MUA will only include these two attributes.  If
-``own_state.prefer_encrypt`` is set to ``mutual``, then the header MUST
+``accounts[A].prefer_encrypt`` is set to ``mutual``, then the header MUST
 have a ``prefer-encrypt`` attribute with the value ``mutual``.
 
 The MUA MUST NOT include more than one valid Level 1 ``Autocrypt``
-header (see :ref:`update-peer-state`).
+header (see :ref:`update-peers`).
 
 If the ``From`` address changes during message composition (e.g., if
 the user selects a different outbound identity), then the MUA MUST
@@ -263,7 +264,7 @@ the following sections for header format definitions and parsing.
 Internal state storage
 ++++++++++++++++++++++
 
-See :ref:`peer-state` for the information stored for each
+See :ref:`peers` for the information stored for each
 communication peer.
 
 Autocrypt MUAs keep state about each peer, to handle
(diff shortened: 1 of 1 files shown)


r10s:
Good idea, looks very good to me.

doc/level1.rst
```diff
@@ -358,9 +359,9 @@ Updating the Autocrypt state for the sending peer depends on:
 If the effective message date is older than the ``last_seen_autocrypt``
 value, then no changes are required, and the update process terminates.
 
-If the Autocrypt header is unavailable, and the effective
-message date is more recent than the current value of ``last_seen``,
-then the peer state should be updated as follows:
+If the Autocrypt header is unavailable, and the effective message date
+is more recent than the current value of ``last_seen``, then the peer
```
dkg: This should be ``peers[A].last_seen``

doc/level1.rst
```diff
@@ -358,9 +359,9 @@ Updating the Autocrypt state for the sending peer depends on:
...
+state should be updated as follows:
```
dkg: "the peer state should be…" → "peers[A] should be…"

doc/level1.rst
```diff
@@ -121,14 +121,15 @@ attributes with ``peer_state[A]``:
...
+.. _accounts:
 
 Accounts controlled by the MUA
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A Level 1 MUA maintains an internal structure ``own_state`` for each
-account on which Autocrypt is enabled. ``own_state`` has the following
-attributes:
+A Level 1 MUA maintains an internal structure ``accounts`` indexed by
+the  the account's :doc:`canonicalized e-mail address
```
nwalfield: "the the"

doc/level1.rst
```diff
@@ -140,11 +141,11 @@ attributes:
...
 If Autocrypt is enabled for a given account, the MUA SHOULD allow the
-user to switch the setting for ``own_state.prefer_encrypt``.  This
+user to switch the setting for ``accounts[A].prefer_encrypt``.  This
 choice might be hidden in something like a "preferences pane".  See
 :ref:`preference-ui` for a specific example of how this could look.
 
-How this information is managed and used is discussed in :ref:`own-state-management`.
+How this information is managed and used is discussed in :ref:`accounts-management`.
 
 .. _peer-state-management:
```
nwalfield: Should peer-state-management also be renamed to peer-management?

doc/level1.rst
```diff
@@ -427,18 +428,18 @@ Recommendations for single-recipient messages
...
+1. If there is no ``peers`` entry, the recommendation is ``disable``.
 2. If there is no ``public_key``, the recommendation is ``disable``.
 3. If the ``public_key`` is known for some reason to be unusable for
    encryption (e.g., it is otherwise known to be revoked or expired),
    then the recommendation is ``disable``.
 4. If the message is composed as a reply to an encrypted message, then
    the recommendation is ``encrypt``.
 5. If both ``state`` is ``mutual`` and
-   ``own_state.prefer_encrypt`` is ``mutual``, then the
+   ``accounts[A].prefer_encrypt`` is ``mutual``, then the
```
dkg: The ``A`` in ``accounts[A]`` is a different ``A`` than the one in ``peers[A]``.  Not sure of the best way to avoid this confusion.  Maybe always use ``peers[P]`` and ``accounts[A]`` ?

dkg:
I think the current proposal here is excellent, and i would approve merging it as-is, though i do have a few comments i've made above.

If @azul doesn't feel like cleaning up the pull request, i'd be happy to merge it as-is, and then i'll make a followup PR that includes the cleanup.

doc/level1.rst
dkg: are you just asking about the anchor label, or the visible text?

if it's the anchor label, it should be ``peers-management`` for symmetry with ``accounts-management``.

If you're asking about the visible text, i'm uneasy with "peer management" because it could be read as implying that you're managing your peers, when what you're managing is state about those peers.

doc/level1.rst
```diff
@@ -294,11 +295,11 @@ state specified here, regardless of what additional state they track.
     a :rfc:`User ID <4880#section-5.11>` that matches the message's
     ``From`` address.
 
-``peer_state[A].state`` semantics
+``peers[A].state`` semantics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``state`` variable of a particular peer's ``peer_state`` data is
-selected from a set range of values:
+The ``state`` value of a particular ``peers`` entry is selected from a
```
nwalfield: The wording is awkward.  I propose:

  The value of the ``state`` attribute can be either:

doc/level1.rst
```diff
@@ -318,15 +319,15 @@ The rough descriptions outlined above are not normative -- they're
...
-.. _update-peer-state:
+.. _update-peers:
 
 Updating Autocrypt Peer State
 +++++++++++++++++++++++++++++
 
-Incoming messages may be processed to update Autocrypt peer state by a
-MUA at receive or display time.
+Incoming messages may be processed to update the ``peers`` entry for the
+sender by a MUA at receive or display time.
```
nwalfield: "an MUA"

doc/level1.rst
```diff
@@ -427,18 +428,18 @@ Recommendations for single-recipient messages
 
 The Autocrypt recommendation for a message composed to a single
 recipient with the e-mail address ``A`` depends primarily on the value
-stored in :ref:`peer_state[A] <peer-state>`. It is derived
+stored in :ref:`peers[A] <peers>`. It is derived
 by the following algorithm:
 
-1. If there is no peer state, the recommendation is ``disable``.
+1. If there is no ``peers`` entry, the recommendation is ``disable``.
```
nwalfield: no entry in ``peers``

dkg:
Talking with @nwalfield further about this addressing business, i think we should index the arrays based on context.  So, during message ingestion, we should talk about ``peers[from-addr]`` and during message composition, we should talk about ``peers[to-addr]`` and ``accounts[from-addr]``.

doc/level1.rst
nwalfield: peers-management is awkward.  I'm using peer-management instead and I changed accounts-management to account-management.

doc/level1.rst
```diff
@@ -318,16 +321,17 @@ The rough descriptions outlined above are not normative -- they're
...
-.. _update-peer-state:
+.. _update-peers:
 
 Updating Autocrypt Peer State
 +++++++++++++++++++++++++++++
 
-Incoming messages may be processed to update Autocrypt peer state by a
-MUA at receive or display time.
+Incoming messages may be processed to update the ``peers`` entry for
+the sender identified by ``sender-addr``, by an MUA at receive or
```
dkg: shouldn't this be ``from-addr`` ?

doc/level1.rst
```diff
@@ -550,9 +556,9 @@ with more than one recipient. These headers MUST be placed in the root
 MIME part of the encrypted message payload. The encrypted payload in
 this case contains one Autocrypt-Gossip header for each recipient,
 which MUST include ``addr`` and ``keydata`` attributes with the
-relevant data from the sender's Autocrypt :ref:`peer state
-<peer-state>` about the recipient. It SHOULD NOT contain a
-``prefer-encrypt`` attribute.
+corresponding values for the recipient identified by ``addr`` as
```
dkg: shouldn't this be ``to-addr`` ?

nwalfield: So, Bob receives a message from Alice, and Carol is in cc.  In this case, Bob probably thinks about his address as being the ``to-addr``.  I think it is okay to use ``addr`` in this case.  Perhaps ``other-addr`` would be better, but I think ``to-addr`` is confusing.

doc/level1.rst
nwalfield: changed.

doc/level1.rst
```diff
@@ -818,9 +824,9 @@ import it to enable Autocrypt.  If the user agrees to do so:
    decryption. See :doc:`bad-import` for more explanation and an
    example.
 
- * If it decrypts the MUA SHOULD import the secret
-   key material as its own Autocrypt (``own_state`` as
-   discussed in :ref:`own-state`).
+ * If it decrypts, then the MUA SHOULD set
+   ``accounts[addr].secret_key`` to the secret key material (as
```
Valodim: this not only affects `secret_key`

doc/level1.rst
```diff
@@ -231,18 +233,20 @@ to an OpenPGP backend for handling).
...
-SHOULD include an Autocrypt header. This header MUST contain the
-corresponding public key material (``own_state.public_key``) as the ``keydata``
-attribute, and the ``sender address`` as the ``addr`` attribute.  The most
-minimal Level 1 compliant MUA will only include these two attributes.  If
-``own_state.prefer_encrypt`` is set to ``mutual``, then the header MUST
-have a ``prefer-encrypt`` attribute with the value ``mutual``.
+During message composition, if the ``From:`` header of the outgoing
+e-mail (the ``from-addr``) matches an address for which the
+Autocrypt-capable MUA has secret key material
+(``accounts[from-addr].secret_key``), it SHOULD include an Autocrypt
```
Valodim: the "it" here sounds like it might refer to the secret key material, I suggest making it an explicit reference to the message

doc/level1.rst
```diff
@@ -318,16 +321,17 @@ The rough descriptions outlined above are not normative -- they're
...
 
-Incoming messages may be processed to update Autocrypt peer state by a
-MUA at receive or display time.
+Incoming messages may be processed to update the ``peers`` entry for
+the sender identified by ``from-addr`` as extracted from the ``From``
+header, by an MUA at receive or display time.
 
-Messages SHOULD be ignored, and the peer state SHOULD NOT be updated in
-the following cases:
+Messages SHOULD be ignored, and ``peers[from-addr]`` SHOULD NOT be
```
Valodim: this SHOULD and SHOULD NOT seems redundant

doc/level1.rst
Valodim: maybe use the role of the address rather than its originating header, as in `rcpt-addr`?

Issue #162, opened 2017-11-04
"Accounts controlled by the MUA" (own-state) should include enabled/disabled?


dkg:
maybe we want to add enabled/disabled to own_state?

PR #206, opened 2017-11-06
add accounts[addr].enabled

dkg:
Making the "enabled" value an explicit value makes it easier to model
and reason about the account.

Closes: #162

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -128,9 +128,11 @@ Accounts controlled by the MUA
 
 A Level 1 MUA maintains an internal structure ``accounts`` indexed by
 the account's :doc:`canonicalized e-mail address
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -140,7 +142,7 @@ Autocrypt is enabled, ``accounts[addr]`` has the following attributes:
    preference for this account.  This is either ``mutual`` or ``nopreference``.
    This SHOULD default to ``nopreference``.
 
-If Autocrypt is enabled for a given account, the MUA SHOULD allow the
+If ``accounts[addr].enabled`` is ``true``, the MUA SHOULD allow the
```
Valodim: Is it necessary to mention that this setting is pointless while autocrypt isn't even enabled?

PR #209, opened 2017-11-06
proposal for new recommendation algorithm

Valodim:


doc/level1.rst
```diff
@@ -432,30 +432,46 @@ Recommendations for single-recipient messages
...
+recommendation:
 - If both ``prefer_encrypt`` and ``accounts[to-addr].prefer_encrypt``
-   are ``mutual``, then the recommendation is ``encrypt``.
+   are ``mutual``, then the preliminary recommendation is ``encrypt``.
 - If ``public_key_timestamp`` is more than a month older than
-  ``last_seen``, the recommendation is ``discourage``.
-- Otherwise, the recommendation is ``available``.
+  ``last_seen``, the preliminary recommendation is ``discourage``.
+- Otherwise, the preliminary recommendation is ``available``.
+
```
nwalfield: Consider the following scenarios, which I believe are not intended:

# last_seen = 1 year ago.
# public_key_timestamp = 1 year + 1 day ago
# no gossip key
# recommendation => available

# last_seen = 1 year ago.
# public_key_timestamp = 1 year + 1 day ago
# gossip_key_timestamp = today
# recommendation => discourage

nwalfield: Another scenario:

We only "reset" if we haven't seen a public key in a month.  Beforehand,
we reset if a single non-Autocrypt message was seen (IIRC).

doc/level1.rst
```diff
@@ -432,26 +415,46 @@ Recommendations for single-recipient messages
...
+value stored in :ref:`peers[to-addr] <peers>`. It is derived using a
+two-phase algorithm.  The first phase computes the preliminary
+recommendation:
+
+The preliminary recommendation is ``disable`` if there is no
+``peers[to-addr]``.
+
+``public_key`` is set to ``peers[to-addr].public_key``.  If
+``public_key`` is revoked, expired, or otherwise known to be unusable
+for encryption, e.g., it is revoked or expired, it is set to null.
```
dkg: revoked, expired, revoked, expired, revoked, expired, revoked, expired, revoked, expired, revoked, expired, revoked, expired,

doc/level1.rst
```diff
@@ -432,26 +415,46 @@ Recommendations for single-recipient messages
...
+
+``gossip_key`` is set to ``peers[to-addr].gossip_key``.  If
+``gossip_key`` is revoked, expired, or otherwise known to be unusable
+for encryption, e.g., it is revoked or expired, it is set to null.
+
+The preliminary recommendation is ``disable``, if both ``public_key``
+and ``gossip_key`` are null.
```
dkg: put the If before the result.

doc/level1.rst
Valodim: :hurtrealbad:

doc/level1.rst
```diff
@@ -432,26 +415,46 @@ Recommendations for single-recipient messages
...
+
+Otherwise, if either ``public_key`` is null, or
+``public_key_timestamp`` is more than a month older than
+``gossip_key_timestamp``, the ``gossip_key`` is used with a
+preliminary recommendation of ``discourage``.
+
+
+Otherwise, the ``public_key`` is used, with the following preliminary
+recommendation:
+- If both ``prefer_encrypt`` and ``accounts[to-addr].prefer_encrypt``
```
dkg: should be ``peers[to-addr].prefer_encrypt`` and ``accounts[from-addr].prefer_encrypt``

doc/level1.rst
```diff
@@ -432,26 +415,46 @@ Recommendations for single-recipient messages
...
+   are ``mutual``, then the preliminary recommendation is ``encrypt``.
+- If ``public_key_timestamp`` is more than a month older than
+  ``last_seen``, the preliminary recommendation is ``discourage``.
```
dkg: This point needs to move above the immediately previous point.  otherwise, we'll choose ``encrypt`` for something that should have been ``reset`` (in the old terminology).

doc/level1.rst
```diff
@@ -565,23 +568,24 @@ MIME part.
...
-An incoming message may contain one or more Autocrypt-Gossip headers
-in the encrypted payload. Each of these headers may update the
+An incoming message may contain one or more ``Autocrypt-Gossip``
+headers in the encrypted payload. Each of these headers may update the
 Autocrypt peer state of the gossiped recipient identified by its
-``addr`` value in the following way:
+``addr`` value (referred to here as ``gossip-addr``) in the following
+way:
+
+1. If the ``gossip-addr`` does not match any recipient in the mail's
```
dkg: remove "the"

b3yond:
Nice PR! I would reccomend to set every "terminates" in "update process terminates" to upper case. This adds to readability :)

doc/level1.rst
```diff
@@ -432,26 +415,47 @@ Recommendations for single-recipient messages
...
+preliminary recommendation of ``discourage``.
+
+
+Otherwise, the ``public_key`` is used, with the following preliminary
+recommendation:
+- If ``autocrypt_timestamp`` is more than a month older than
+  ``last_seen``, the preliminary recommendation is ``discourage``.
+- If both ``peers[to-addr].prefer_encrypt`` and
+   ``accounts[from-addr].prefer_encrypt`` are ``mutual``, then the
+   preliminary recommendation is ``encrypt``.
```
azul: I would like to move this aspect into the second step. That way the outcome of the first step can be one of:
* available
* discourage
* disable

The second part would then switch on encryption in certain conditions.

azul:
Looks good to me now. Also does not look so much more complicated anymore now.

doc/level1.rst
azul: #211 has this for comparison.
I'm happy with this either way.

doc/level1.rst
```diff
@@ -107,13 +107,26 @@ following attributes with ``peers[addr]``:
...
+  processed from this peer.
+* ``autocrypt_timestamp``: The UTC timestamp of the most recent
+  effective date of all messages containing a valid ``Autocrypt`` header
+  that the MUA has processed from this peer.
+* ``public_key``: The value of the ``keydata`` attribute derived from
+  the most recent ``Autocrypt`` header received from the peer.
+* ``prefer_encrypt``: The ``prefer-encrypt`` value (either
+  ``nopreference`` or ``mutual``) derived from most recent ``Autocrypt``
+  header received from the peer.
+
```
nwalfield: nit: it's not the most recently seen Autocrypt header, but the youngest Autocrypt header that has ever been seen for the given peer.

doc/level1.rst
```diff
@@ -432,26 +415,48 @@ Recommendations for single-recipient messages
...
+``gossip_key`` is set to ``peers[to-addr].gossip_key``.  If
+``gossip_key`` is revoked, expired, or otherwise known to be unusable
+for encryption, it is set to null.
+
+If both ``public_key`` and ``gossip_key`` are null, the preliminary
+recommendation is ``disable``.
+
+Otherwise, if either ``public_key`` is null, or
+``autocrypt_timestamp`` is more than a month older than
+``gossip_key_timestamp``, the ``gossip_key`` is used with a
```
nwalfield: To avoid ambiguity, can we say 4 weeks here instead of one month.

doc/level1.rst
```diff
@@ -432,26 +415,48 @@ Recommendations for single-recipient messages
...
+preliminary recommendation of ``discourage``.
+
+
+Otherwise, the ``public_key`` is used, with the following preliminary
+recommendation:
+- If ``autocrypt_timestamp`` is more than a month older than
+  ``last_seen``, the preliminary recommendation is ``discourage``.
+- Otherwise, the preliminary recommendation is ``available``.
+
```
nwalfield: I like this wording, and I think it fixes the problem that (gossip, discourage) could have been preferred to (public, available).  Good job.

b3yond: but it should also say 4 weeks of course^^

doc/level1.rst
dkg: I prefer 35 days over 4 weeks.  There are many systems and social structures that actually send mail once per month (e.g. account status updates, scheduled checkins), and it would be good if scheduled monthly communications don't fall over the duration gap.  the `recommendation-cleanup` branch (see #215) contains a 35-day window.

azul:
I think we should merge this now...

Valodim:
I wonder what happens to the PR that wants to merge into this branch, when we merge this? Let's go for it!

dkg:
@valodim, that PR is already merged in here.

Valodim:
Most approved PR we've ever had I AM PUSHING THE BUTTON

dkg:
\o/

Valodim:
This change went more smoothly than it had any right to in these last minute circumstances. Great job everyone!

b3yond:
Nice! That's it with the milestone, too!

PR #215, opened 2017-11-06
break up recommendation algorithm and explain more.

dkg:
* clarify that the recommendation algorithm returns both a
   recommendation for the behavior of the encryption UI, and a list of
   keys.

 * move the "disabled" logic above the preliminary recommendation

 * add section headers to the phases of the algorithm

 * remove the ``public_key`` and ``gossip_key`` intermediate
   variables, which seem to add more confusion than they remove, since
   they shadow existing variables with the same name.

 * make preliminary recommendation an explicit intermediate variable.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -446,14 +446,14 @@ Preliminary Recommendation
 __________________________
 
 If either ``public_key`` is ``null``, or ``autocrypt_timestamp`` is
-more than a month older than ``gossip_key_timestamp``, set
+more than 35 days older than ``gossip_key_timestamp``, set
 ``target-keys[to-addr]`` to ``gossip_key`` and set
 ``preliminary-recommendation`` to ``discourage`` and skip to the
 :ref:`final-recommendation-phase`.
 
 Otherwise, set ``target-keys[to-addr]`` to ``public_key``.
 
-If ``autocrypt_timestamp`` is more than a month older than
+If ``autocrypt_timestamp`` is more than 35 days older than
 ``last_seen``, set ``preliminary-recommendation`` to ``discourage``.
 
 Otherwise, set ``preliminary-recommendation`` to ``available``.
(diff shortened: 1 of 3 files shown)


doc/level1.rst
```diff
@@ -416,72 +422,91 @@ Recommendations for single-recipient messages
...
+recipient ``target-keys``.
+
+The aggregate ``ui-recommendation`` for the message is derived as
+follows:
 
-1. If any recipient has a recommendation of ``disable``, then the
-   message recommendation is ``disable``.
+1. If any recipient has a ``ui-recommendation`` of ``disable``, then
+   the message's ``ui-recommendation`` is ``disable``.
 2. If the message being composed is a reply to an encrypted message,
```
Valodim: This is redundant, replies will always be "disable" or "encrypt". might be fine to have it regardless, but it shouldn't be necessary if we didn't screw up above

doc/level1.rst
```diff
@@ -393,7 +393,13 @@ The Autocrypt recommendation depends on the recipient
...
-Autocrypt can produce four possible recommendations:
+The output of the Autocrypt recommendation algorithm has two elements:
+
+ * ``ui-recommendation``: a single state recommending the state of the
+   encryption user interface, described below.
+ * ``target-keys``: a map of recipient addresses to public keys.
+
+``ui-recommendation`` can take four possible values:
 
  * ``disable``: Disable or hide any UI that would allow the user to
```
Valodim: We could mention here that this recommendation happens exactly iff encrypting is in fact not available.

dkg:
thanks, i think i've resolved both of these suggestions

b3yond:
Oops, didn't see this and commited https://github.com/autocrypt/autocrypt/pull/209/commits/fc7570a1ed83cf51591ef88d16684ef43efe985d directly

Valodim:
removed fc7570a1ed83cf51591ef88d16684ef43efe985d via force push in that PR to resolve conflict here

Issue #77, opened 2017-06-11
"scan for" vs. "wait for" for autocrypt setup message import


dkg:
if we say "wait for" then the transfer process requires starting on the new device.

if we say "scan for" then we require clients to be able to scan folders

comradekingu:
Is there anything wrong with "look for"

azul:
can we scan for if possible and if not or we find nothing we wait for?

PR #216, opened 2017-11-06
Clarify different ways of finding Autocrypt Setup Message.

dkg:
Provide a couple of specific example suggestions for finding the
Autocrypt Setup Message, without suggesting that the MUA has to be
able to do them all.

Closes: #77


azul:
looks good to me.

PR #220, opened 2017-11-07
Add gossip headers example

### LINKED ISSUE #219: Add example for a message with key gossip

juga0:
Should fix #219

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -608,6 +608,7 @@ The ``Autocrypt-Gossip`` header has the format as the ``Autocrypt``
 header (see `autocryptheaderformat`_). Its ``addr`` attribute
 indicates the recipient address this header is valid for as usual, but
 may relate to any recipient in the ``To`` or ``Cc`` header.
+See example in :ref:`autocrypt-gossip-example`
 
 Key Gossip Injection in Outbound Mail
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -1026,6 +1027,229 @@ Example Copy when a Reply can't be Encrypted
...
+Alice send an e-mail to Bob and Carol with Autocrypt Gossip headers.
+
+::
+
+    Delivered-To: <bob@autocrypt.example>
+    From: Alice <alice@autocrypt.example>
+    To: Bob <bob@autocrypt.example>, Carol <carol@autocrypt.example>
+    Subject: an Autocrypt Gossip header example
+    Autocrypt: addr=alice@autocrypt.example; keydata=
+      mQGNBFn+L+YBDAC3jsOXmFKwKfUh/WxaOErSMMdL1NJzzFCDf4oo0XD5b4ldfVGP09PsNXg5bzUW
```
Valodim: this keydata seems significantly shorter than the one below. is that just 2048 bits?

dkg: it is copied from the e-mail example above.

inspecting them both suggests that the keydata is truncated somehow :(

dkg: I'll update both of them, replacing them with the public key derived from the Autocrypt Setup Message.

Valodim: :+1:

dkg: Also, i think the gossip message needs to be encrypted to Alice, Bob, and Carol, not just to Bob and Carol, since iirc, we recommend "encrypt to self".  That will also make it decryptable just from the setup message example.

@juga0 volunteered to work on this.

dkg:
I'm closing this pull request because it is superceded by #222

PR #224, opened 2017-11-08
 change reasoning around recommendation and how MUAs should t…

hpk42:
…reat it. Also drop the reference to "other-crypto-interop.rst" which is an unfinished and not too useful doc -- which is irritating if it is referenced from the spec. I don't think it's neccessary for 1.0.0 and can be easily added back for 1.0.1 if needed.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -417,10 +417,10 @@ Autocrypt recommendation.
 
 An Autocrypt-capable MUA may have other means for making this
 decision outside of Autocrypt. However, ignoring the recommendation
-algorithm may lead to unexpected unreadable mail which leads to
-user frustration.  In particular, encrypting by default may lead
-to users disabling encryption support and thus lowering the
-the overall amount of encrypted e-mail.
+algorithm may lead to an increase in unexpected unreadable mail which
+frustrates users.  In particular, simply encrypting when technically possible
+may lead to users disabling encryption support and thus lowering the overall
+amount of encrypted e-mail.
 
 The Autocrypt recommendation depends on the recipient
 addresses of the draft message.  When the user changes the
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -412,15 +412,15 @@ Provide a recommendation for message encryption
 On message composition, an Autocrypt-capable MUA
 can decide whether to try to encrypt the new e-mail
 message.  Autocrypt provides a recommendation for the MUA.
+All Autocrypt-capable MUAs should be able to calculate the same
+Autocrypt recommendation.
 
-Any Autocrypt-capable MUA may have other means for making this
-decision outside of Autocrypt (see :doc:`other-crypto-interop`).
-Autocrypt provides a recommendation, but there is no
-requirement for Autocrypt-capable MUAs to follow this
-recommendation.
-
-That said, all Autocrypt-capable MUAs should be able to calculate
-the same Autocrypt recommendation.
+An Autocrypt-capable MUA may have other means for making this
+decision outside of Autocrypt. However, ignoring the recommendation
+algorithm may lead to unexpected unreadable mail which leads to
+user frustration.  In particular, encrypting by default may lead
+to users disabling encryption support and thus lowering the
+the overall amount of encrypted e-mail.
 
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -412,15 +412,15 @@ Provide a recommendation for message encryption
...
-That said, all Autocrypt-capable MUAs should be able to calculate
-the same Autocrypt recommendation.
+All Autocrypt-capable MUAs should be able to calculate the same
+Autocrypt recommendation.
+
+An Autocrypt-capable MUA may have other means for making this
+decision outside of Autocrypt. However, ignoring the recommendation
+algorithm may lead to unexpected unreadable mail which leads to
+user frustration.  In particular, encrypting by default may lead
+to users disabling encryption support and thus lowering the
```
Valodim: The the

doc/level1.rst
```diff
@@ -412,15 +412,15 @@ Provide a recommendation for message encryption
...
-requirement for Autocrypt-capable MUAs to follow this
-recommendation.
-
-That said, all Autocrypt-capable MUAs should be able to calculate
-the same Autocrypt recommendation.
+All Autocrypt-capable MUAs should be able to calculate the same
+Autocrypt recommendation.
+
+An Autocrypt-capable MUA may have other means for making this
+decision outside of Autocrypt. However, ignoring the recommendation
```
Valodim: I'd say "an increase", as our algorithm surely doesn't prevent unreadable mail entirely

doc/level1.rst
```diff
@@ -412,15 +412,15 @@ Provide a recommendation for message encryption
 On message composition, an Autocrypt-capable MUA
 can decide whether to try to encrypt the new e-mail
 message.  Autocrypt provides a recommendation for the MUA.
-
-Any Autocrypt-capable MUA may have other means for making this
-decision outside of Autocrypt (see :doc:`other-crypto-interop`).
```
Valodim: :+1:

dkg:
This change backs off from the "meh, MUAs don't really have to follow the recommendation" language, which i agree was far too weak.  The work we've put into defining a reasonable recommendation should be used :)

That said, i think this change goes further than it needs to, by identifying aggressive encryption policies in particular as the "bad" thing that happens from not following the recommendation.

there are lots of ways to not-follow the Autocrypt recommendation algorithm, and not all of them result in aggressive encryption.  some of them are at least as bad (e.g., mailing cleartext replies to an encrypted message, with quoted and attributed text in the clear).

I don't see why we should be calling out one way to fail to follow the encryption algorithm, and not call out another.

IIRC, the intent of the initial paragraph was to permit clients to claim "Autocrypt" compatibility, while still experimenting with (A) new data sources or statistical metrics, and (B) new recommendation algorithms while still trying to minimize/avoid the unnecessary pain points typically inflicted on users by encrypted mail.

I'd be happy to see the paragraph cleaned up in line with that general sense, but i think it's a mistake to insert an anti-aggressive-encryption paragraph here.  We already have one of those, front and center in the Overview, section 1.1.

Valodim:
Sounds reasonable to me. Do you want to propose an alternative change?

dkg:
Sure, i've just proposed #227 as an alternative that tries to capture the sentiment in the discussion above.

Obviously, if we merge #224 we should close #227 as wontfix, and vice versa.  Either one would be an improvement on the status quo, imho.

lefherz:
Is there a reason #224 and #227 are not in the Level 1 Milestone?

dkg:
good catch.

hpk42:
My guess is that the spec will be read a lot by folks who have a strong existing-pgp-practise informed view. And i claim there is a "more encryption is better" principle often invoked.  Some feedback i got from George (i am asking him if i can forward it to the ML) also indicates this.  Our evolved approach to consider the ecosystem as a whole, not just a single mail transaction, is not a common one.  Therefore, refering back to it even more than one time is ok -- it reminds the reader of the overall goal.  The spec is not just there for technical clarity IMHO but also to give hints / provide brief background  for our choices.

hpk42:
Merged #227 in favour of my own one.

PR #227, opened 2017-11-09
Re-frame introduction to the Autocrypt Recommendation

dkg:
This is meant to be considered as an alternative to PR #224.

As i wrote in the discussion on that thread:

  IIRC, the intent of the initial paragraph was to permit clients to
  claim "Autocrypt" compatibility, while still experimenting with (A)
  new data sources or statistical metrics, and (B) new recommendation
  algorithms while still trying to minimize/avoid the unnecessary pain
  points typically inflicted on users by encrypted mail.

Hopefully this captures that intent better.

With the increased text here, i thought it was better to break out the
technical description of the structure of the Recommendation algorithm
(its inputs and outputs) into its own subsection, so that the
descriptive/normative/motivational text is distinct from the technical
specifics.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -413,18 +413,34 @@ On message composition, an Autocrypt-capable MUA
 can decide whether to try to encrypt the new e-mail
 message.  Autocrypt provides a recommendation for the MUA.
 
-Any Autocrypt-capable MUA may have other means for making this
-decision outside of Autocrypt (see :doc:`other-crypto-interop`).
-Autocrypt provides a recommendation, but there is no
-requirement for Autocrypt-capable MUAs to follow this
-recommendation.
+All Autocrypt-capable MUAs should be able to calculate the same
+Autocrypt recommendation.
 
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -413,18 +413,34 @@ On message composition, an Autocrypt-capable MUA
...
+including:
+
+ * does it increase the chance of producing unexpectedly unreadable
+   mail (for either the sender or the recipient)?
+ * does it leak previously encrypted content in the clear?
+ * does it force the user to confront a choice they do not have the
+   information or knowledge to make safely?
+
+If an implementation deviates from the Autocrypt recommendation in a
+meaningful and useful way, the implementer should describe the
```
hpk42: i'd make this a SHOULD.

dkg: SHOULD is usually intended for what implementations are expected to do.  This sentence is about what *implementers* are expected to do.  I'd rather not mix them up, but i don't feel strongly about it.  If you want to add a patch to this series that makes this a SHOULD i would be fine with it.

Issue #225, opened 2017-11-08
drop or fix content of linked ":doc:"s ...


hpk42:
level1.rst contains a number of ``:doc:`` references that are not part of the spec and will not be readable/clickable from the PDF.   For each link i suggest:

- to drop them if they are not really neccessary to understanding level 1 ... or 

- to try incorporate them directly in the spec (e.g. canonicalized e-mail address) 

- if they remain, then make sure at least the directly linked doc has comparable quality
  to the spec

Valodim:
Yes! I very much agree with this, and it should be part of the level 1 freeze. It's very important to have a self-contained spec (whether it's pdf or something else)

dkg:
I agree with this as well.  more to do…

dkg:
So i see five external `:doc:` references remaining:

 * in the introduction, where it calls out to `next-steps`
     I don't mind leaving this one in (though i'm not sure it should talk specifically about "Level 2", but maybe should talk about "future revisions").
 * two references to `address-canonicalization`
     This should probably be incorporated explicitly, or else we should drop the idea that Autocrypt clients all do the same address canonicalization.  I think if we want predictability, we should encourage address canonicalization, though.
 * `optional-state`
     This is now out of date given the overhaul to the `peers` table, but it's could be brought back into shape.  I'm reluctant to cut out all references to it from the spec because i want to honor the work that @BjarniRunar put into it, and to encourage more experimentation like this.  I'm not sure how to resolve this one.  Can we designate a URL within https://autocrypt.org/ as a location for this kind of experimentation and forward-looking proposals that aren't yet merged into the main spec?
 * `other-crypto-interop`
     This will (hopefully) be killed by either #224 or #227
 * `bad-import`
     This is concretely useful information about a specific failure mode of a common implementation.  Maybe it also deserves a permanent home somewhere that we can point to as an external URL in the spec?

It might be useful to open a separate issue for each one of these remaining concerns.

dkg:
OK, we have 3 left, and i've opened separate tickets for each of them, along with  my general suggestions for what should be done:

- [ ] #237 
- [ ] #238 
- [ ] #239

hpk42:
There now only is #238 left for removing the last occurence of ``:doc:`` (optional-state).  Close this issue in favor of #238.

PR #229, opened 2017-11-10
rewrite the very intro to explain higher level focus of autocrypt and…

hpk42:
… how it differs from past ones. After hearing some first comments from fresh readers i think it's important to better scope/motivate what we are doing first and also directly name major differences.   This also makes it easy to remain motivated to read things when you are coming from a (very) "pgp-skeptical" experience. This PR also strikes the reference to "next steps" which is a very drafty document (addresses #225).

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -6,51 +6,43 @@
 Autocrypt Level 1: Enabling encryption, avoiding annoyances
 ===========================================================
 
-This specification is the result of maybe 2000 hours of talking, testing and
-coding iterations during 2017. Around 20 developers and many more users and
-trainers participated overall in Autocrypt sessions. The majority of these
-participants have engaged with the e-mail encryption space for a
-long time. There was overwhelming agreement that fresh approaches are needed
-and particularly ones that address usability concerns.
+Autocrypt aims to incrementally and carefully replace cleartext e-mail
+with end-to-end encrypted e-mail. This differs from the traditional focus
+on maximizing the security of an individual mail communication.
+**Sometimes Autocrypt recommends to send cleartext mail even though
+encryption would seem technically possible.** This is because we want to
+avoid unreadable mail for users. Users may mix both Autocrypt-capable
+and traditional mail apps and they may loose devices or otherwise loose
+the ability to decrypt in unrecoverable ways.  Reverting to cleartext
+when we suspect such situations is a key part of our aim to stay out of
+the way of users.
 
-Autocrypt aims to incrementally replace cleartext e-mail with end-to-end
-encrypted e-mail. It aims to sustainably grow the overall percentage of
-end-to-end encrypted messages relayed via the e-mail provider network.
-This differs from the traditional focus on maximizing the security of an
-individual mail communication. **Sometimes Autocrypt recommends to
-send cleartext mail even though encryption would be technically
-possible.** This is because of well-known annoyances around "i can't
-read your encrypted mail" situations. Autocrypt operates opportunistically
-and tries to stay out of the way of users who want to, first of all,
-communicate.
+Another major difference in approach is that Autocrypt Level 1 only
+defends against passive data collection attacks. We share and support
+:rfc:`the new perspective stated in RFC7435 ("Opportunistic Security: Some
+Protection Most of the Time") <7435#section-1.2>`. Protection against
+active adversaries (those which modify messages in transit) is the aim
+of future specifications.
 
-Another difference to traditional approaches is that Autocrypt Level 1
-only defends against passive data collection. Protection against active
-adversaries (those which modify messages in transit) is the aim of
-future specifications. We share and support :rfc:`the new perspective
-stated in RFC7435 ("Opportunistic Security: Some Protection Most of the
-Time") <7435#section-1.2>`.
+**Level 1 makes it easy for users to encrypt, based on an automatic and
+decentralized key distribution mechanism. There are no dependencies on
+key servers and it is meant to work with existing e-mail providers.**
+Level 1 focuses on the use of Autocrypt on a single device. Users get
+rudimentary support on using Autocrypt on more than one device or mail app.
+This is internally realized through sending and receiving an Autocrypt
+Setup Message, secured by manually entering a long number. Improving
+usability for maintaining synchronized Autocrypt state on multiple
+devices is the aim of future specification efforts.
 
-The Level 1 specification makes it easy for users to encrypt. There are
-no changes required from e-mail providers and there are no dependencies
-on key servers. The spec describes the basic capabilities required for a
-mail app to be Autocrypt-capable at Level 1, allowing it to exchange
-end-to-end encrypted e-mails with other Autocrypt-capable mail apps. For
-those familiar with existing e-mail specs, we use the term "mail app"
-and "MUA" interchangeably throughout the doc.
-
-For ease of implementation and deployment, Level 1 focuses on the use of
-Autocrypt on a single device. However, users get support for transfering
-Autocrypt secret keys to other devices. They need to enter a
-long number for securing the transfer which is technically realized
-through sending and receiving an e-mail message. We aim to improve
-usability around keeping synchronized Autocrypt state on multiple
-devices in subsequent Autocrypt specifications.
-
-We designed Level 1 to be relatively easy for developers to adopt. There is
-detailed guidance on protocol, internal state and user interface concerns.
-We have a good track record of supporting new implementers. Please
-don't hesitate to ask back or send comments.
+**Last but not least, Level 1 is meant to be relatively easy for
+developers to adopt.** It describes the basic capabilities required for
+a mail app to be Autocrypt-capable at Level 1, allowing it to exchange
+end-to-end encrypted e-mails with other Autocrypt-capable mail apps. The
+spec contains detailed guidance on protocol, internal state and user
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -6,22 +6,40 @@
...
+Time") <7435#section-1.2>`.
 
 For ease of implementation and deployment, Level 1 focuses on the use
-of Autocrypt on a single device.  We intend to :doc:`support
-multi-device synchronization (and other features) as part of Level
-2<next-steps>`.  We want to keep Level 1 simple enough that it's easy
-for developers to adopt it so we can drive efforts from real-life
-experiences as soon as possible.
+of Autocrypt on a single device. It also defines an initial mechanism for
+sharing secret keys between mail apps of a single user. We aim to improve the
```
Valodim: ..."without introducing dependencies on more than very basic MUA capabilities"?

doc/level1.rst
```diff
@@ -6,22 +6,42 @@
...
+encrypted e-mail. The Level 1 specification makes it easy for users to
+encrypt. There are no changes required from e-mail providers and there
+are no dependencies on key servers. The spec describes the basic
+capabilities required for a mail app to be Autocrypt-capable at Level 1,
+allowing it to exchange end-to-end encrypted e-mails with other
+Autocrypt-capable mail apps.
+
+This specification is the result of maybe 2000 hours of talking and
+coding iterations during 2017. Around 20 people and many more users and
+trainers participated in Autocrypt sessions. The majority of these
```
dkg: This seems to imply that users and trainers are not people.  I think that goes against the emphasis of the project :)

doc/level1.rst
```diff
@@ -6,22 +6,42 @@
...
+participants have engaged with the e-mail encryption space for a
+long time. There was overwhelming agreement that fresh approaches focusing
+on usability concerns are needed.
+
+Autocrypt aims to sustainably grow the overall percentage of end-to-end
+encrypted messages relayed via the e-email provider network. This
```
dkg: e-email

doc/level1.rst
```diff
@@ -6,22 +6,42 @@
...
+future specifications. We share and support :rfc:`the new perspective
+stated in RFC7435 ("Opportunistic Security: Some Protection Most of the
+Time") <7435#section-1.2>`.
+
+For ease of implementation and deployment, Level 1 focuses on the use of
+Autocrypt on a single device. Users get basic support for transfering
+secret keys between Autocrypt-capable mail apps. They need to enter a
+long number for securing the transfer, technically realized through
+sending and receiving an e-mail message. We aim to improve the usability
+around secret key sharing in subsequent Autocrypt specifications.
```
dkg: I hope that future autocrypt specifications will aim to improve the multi-device use case -- not just secret key sharing.  we should stay focused on the user concerns; key management is a means to an end.

doc/level1.rst
```diff
@@ -6,22 +6,42 @@
...
+
+We designed Level 1 to be relatively easy for developers to adopt. There is
+detailed guidance on protocol, internal state and usability issues.
+We have a good track record of supporting new implementers. Please
+don't hesitate to ask back and send comments.
```
dkg: native en-US nit-pick: "ask back" feels weird and awkward.  How about:

Please don't hesitate to [contact the project](https://autocrypt.org/en/latest/contact.html) with comments, questions, and suggestions.

doc/level1.rst
```diff
@@ -6,22 +6,42 @@
 Autocrypt Level 1: Enabling encryption, avoiding annoyances
 ===========================================================
 
-Autocrypt makes it easy for people to encrypt email.  This document
-describes the basic capabilities required for a mail app to be
-Autocrypt-capable at Level 1.
-
-The design of Level 1 is driven by usability concerns and by the
-realities of incremental deployment. A user may mix both
-Autocrypt-enabled MUAs and traditional MUAs, and we'd
```
dkg: this update seems to have lost this "may mix both Autocrypt-enabled MUAs and traditional MUAs" idea.  do we not need to express it any more?

dkg:
Despite a few nit-picks above, i like this change overall.  If the nit-picks were resolved, i'd approve it.

I note that it seems to be trying to sneak in the phrase "Mail app" to the introduction instead of using "MUA" as the rest of the document does, but i can live with that.

doc/level1.rst
```diff
@@ -6,22 +6,51 @@
...
-supporting users who want to explicitly turn on encryption.
-
-For ease of implementation and deployment, Level 1 focuses on the use
-of Autocrypt on a single device.  We intend to :doc:`support
-multi-device synchronization (and other features) as part of Level
-2<next-steps>`.  We want to keep Level 1 simple enough that it's easy
-for developers to adopt it so we can drive efforts from real-life
-experiences as soon as possible.
+This specification is the result of maybe 2000 hours of talking, testing and
+coding iterations during 2017. Around 20 developers and many more users and
```
Valodim: ...20 developers?

doc/level1.rst
```diff
@@ -6,22 +6,51 @@
...
+trainers participated overall in Autocrypt sessions. The majority of these
+participants have engaged with the e-mail encryption space for a
+long time. There was overwhelming agreement that fresh approaches are needed
+and particularly ones that address usability concerns.
+
+Autocrypt aims to incrementally replace cleartext e-mail with end-to-end
+encrypted e-mail. It aims to sustainably grow the overall percentage of
```
Valodim: maybe switch out the second "aims" here for an "attempts"

doc/level1.rst
```diff
@@ -6,22 +6,51 @@
...
+stated in RFC7435 ("Opportunistic Security: Some Protection Most of the
+Time") <7435#section-1.2>`.
+
+The Level 1 specification makes it easy for users to encrypt. There are
+no changes required from e-mail providers and there are no dependencies
+on key servers. The spec describes the basic capabilities required for a
+mail app to be Autocrypt-capable at Level 1, allowing it to exchange
+end-to-end encrypted e-mails with other Autocrypt-capable mail apps. For
+those familiar with existing e-mail specs, we use the term "mail app"
+and "MUA" interchangeably throughout the doc.
```
Valodim: Is this necessary? We never use mail app anywhere else in the spec so far. These terms are also from the opposite spectrum of "specialized tech term" to "new and popular term in informal language", which can lead to unnecessary confusion. I'd prefer if we stuck to one term consistently.

doc/level1.rst
```diff
@@ -6,22 +6,51 @@
...
-multi-device synchronization (and other features) as part of Level
-2<next-steps>`.  We want to keep Level 1 simple enough that it's easy
-for developers to adopt it so we can drive efforts from real-life
-experiences as soon as possible.
+This specification is the result of maybe 2000 hours of talking, testing and
+coding iterations during 2017. Around 20 developers and many more users and
+trainers participated overall in Autocrypt sessions. The majority of these
+participants have engaged with the e-mail encryption space for a
+long time. There was overwhelming agreement that fresh approaches are needed
+and particularly ones that address usability concerns.
```
dkg: Leading with this paragraph feels super defensive to me.  I don't think we need to be defensive.

The first paragraph ought to be "what is this thing"

If we really need a defensive paragraph, it should come at the end of this section.

doc/level1.rst
hpk42: added it back with thenew commit.

doc/level1.rst
hpk42: i completely stroke it for now -- i don't see it as defensive, rather as providing context of how the spec evolved to be what it is. Otherwise "autocrypt team" remains rather abstract.

doc/level1.rst
hpk42: i was previsouly under the impression that we used mail app throughout the spec.  That isn't the case (anymore?).  I am fine with using MUA in the technical sections but i'd like to use a language in the very intro that does not alienate all the young folks who have never heart of MUAs. And i don't want to use the first paragraph to introduce/explain MUA.

hpk42:
i rewroked the intro once more to accomodate all comments.  I think it is not clearer sectioned -- first addressing the terrifying prospect of dropping down to cleartext.  Then talking about the "new perspective" (defend against passive adversaries), then the "making it easy for users to encrypt" and lastly making it easy for developers to adopt with an invitation to join efforts.  Hope you find it all agreeable. I'd like to user-test the spec with 1-2 mor folks i want to contact for review :)

doc/level1.rst
```diff
@@ -6,22 +6,43 @@
...
-2<next-steps>`.  We want to keep Level 1 simple enough that it's easy
-for developers to adopt it so we can drive efforts from real-life
-experiences as soon as possible.
+Autocrypt aims to incrementally and carefully replace cleartext e-mail
+with end-to-end encrypted e-mail. This differs from the traditional focus
+on maximizing the security of an individual mail communication.
+**Sometimes Autocrypt recommends to send cleartext mail even though
+encryption would seem technically possible.** This is because we want to
+avoid unreadable mail for users. Users may mix both Autocrypt-capable
+and traditional mail apps and they may loose devices or otherwise loose
```
Valodim: lose

Valodim:
looks good to me! :+1:

dkg:
I've just pushed a fix to the "ask back" wording.  I now think this branch is ready to be merged.

PR #234, opened 2017-11-11
Add numbering to subsections

dkg:
This should make it easier to refer to subsections of the document
specifically.


hpk42:
great, thanks! f and me struggled with this during the sprint for half an hour and failed!

Issue #239, opened 2017-11-12
move `doc/bad-import.rst` to somewhere else, reference from spec


dkg:
`doc/bad-import.rst` is concretely useful information about a specific failure mode of a common implementation (GnuPG).  It's also somewhat out of date.

It deserves a permanent home somewhere, and we should point it as an external URL in the `level1` spec.

Since it's GnuPG-specific, i think the permanent home should be GnuPG-specific too.  What about using https://dev.gnupg.org/T3277 ?

Blocks: #225

hpk42:
i think it's fine to substitute the ``:doc:bad-import`` with the link to https://dev.gnupg.org/T3277 now and close this issue and remove the bad-import file.

Issue #88, opened 2017-06-11
Format bad-import file that documents preventing private key injection


azul:
Currently not formated yet.

dkg:
PR #240 should also close this issue.

dkg:
This was closed when #240 was merged.

PR #240, opened 2017-11-12
Refer to GnuPG bug report instead of tracking it in Autocrypt

dkg:
We've identified a problem with GnuPG importing
symmetrically-encrypted data safely based on the passphrase.

This should be reported on the GnuPG bug-tracker, not held in the
Autocrypt repo. I've pushed as much detail as i think is relevant on
this problem to the GnuPG bug-tracker itself, so i think we can just
point there.

Closes: #239, #88


doc/level1.rst
```diff
@@ -928,12 +928,12 @@ import it to enable Autocrypt.  If the user agrees to do so:
...
 
- * The MUA should try decrypting the message with the supplied
-   Setup Code.  The Code serves both for decryption as well as
+ * The MUA should try decrypting the message with the supplied Setup
+   Code.  The Code serves both for decryption as well as
    authenticating the message.  Extra care needs to be taken with some
    PGP implementations that the Setup Code is actually used for
-   decryption. See :doc:`bad-import` for more explanation and an
-   example.
+   decryption. For example, `this is quite tricky to do correctly with
```
juga0: tricky? :)

dkg: yep, tricky -- we could use the word "difficult" instead if we want more formal language.

dkg: amended to use "difficult" instead of "quite tricky"  -- thanks for the language check!

Issue #236, opened 2017-11-11
Maintain knowledge about time of original receipt when sending out gossip keys


hpk42:
Raised by George Danezis who reviewed the Nov7th version of our 1.0.0rc1 spec:

> In 2.6 Key gossip: there is a general issue with key gossip. The
> sender of the gossip header does not specify how recent or stale the
> gossip information is, and the recipient MUA uses the messages
> receiption data as a proxy for this. This means that when I send my
> yearly Holliday greetings card to family, that I contact once a year,
> I may be propagating stae keys to all my  contacts. Is that an issue? 
> Should the gossip only include "recent" keys? Should it include a 
> timestamp of how recent the info is? (At the expense of contact privacy though).

I agree with the concern.  How about we introduce a "seenfirst_timestamp" which is the time someone got a mail of a key that is send further as a gossip key?  Gossiping a Gossip key further can then maintain that timestamp.  If we gossip gossip keys, contact privacy is less of an issue. we could say that the timestamp should only maintain a time granularity of one week or so to reduce risks oif contact privacy leakage.

Valodim:
The main goal of gossip is to always be able to "reply all", and gossip keys are only sent inside encrypted messages. So if the key used for some recipient is stale, then the first message will have been unreadable to them as well. The entire thread will remain unreadable, but this situation will remedy itself once the affected person replies (to all), propagating their new autocrypt key.

A related issue is that if we receive a gossip key that differs from one we received directly from the person. In that case, one of them will likely be wrong. We considered this before and couldn't come up with a good answer, so we said we'd leave it as is for level 1, iirc.

dkg:
I thnk we're talking about Alice sending mail to Bob and Carol, with both Bob and Carol's keys in the gossip header.  I'll write below from Bob's perspective -- learning about what Alice thinks of Carol's key. 

If Alice is willing to mail Carol using this key, right now, then at some level Bob understands that Alice believes this key is "correct right now".  That's all that Bob should learn about Alice's internal state from this transaction (which he's already learned).

I don't understand the advantage of introducing a "seenfirst_timestamp" as proposed by @hpk42 above.  I don't see how it folds into any decision-making algorithm, and i don't see how Bob would adjudicate between two different `seenfirst_timestamp` values (e.g. if he receives a message from David to him and Carol).

Keep it simple.  If we get into transitive disclosures of date of key visibility, we're into a tangle that we don't want to get into, certainly for level 1.

[...]

PR #243, opened 2017-11-12
Fix/improve Gossip keys

hpk42:
This fixes issues raised in #236 : 

- Make sure we gossip the same key as we are using to encrypt to a recipient, in any given outbound message. 

- record original receive time of a peer's Autocrypt key (through whatever mediator) 
  in the new and required ``origin-timestamp`` attribute of ``Autocrypt-Gossip`` headers. 
  This allows us to have meaningful comparisons between ``autocrypt_timestamp`` and
  ``gossip_timestamp`` which now refer to the same basic event (receiving a mail
  from a peer about her own autocrypt key).

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -656,11 +656,13 @@ for a gossip header in :ref:`autocrypt-gossip-example`).
 Key Gossip Injection in Outbound Mail
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Any ``Autocrypt-Gossip`` headers MUST be placed in the root
+A MUA MAY include ``Autocrypt-Gossip`` headers in messages
+with more than one recipient. These headers MUST be placed in the root
 MIME part of the encrypted message payload. The encrypted payload in
 this case contains one ``Autocrypt-Gossip`` header for each recipient,
 each of which MUST contain the ``addr``, ``keydata`` and the
-gossip-header specific ``origin-timestamp`` attributes.
+gossip-header specific ``origin-timestamp`` attributes. It SHOULD
+NOT contain the ``prefer-encrypt`` attribute.
 
 A MUA MUST gossip the same public key for a recipient that
 it uses to encrypt to that recipient. The values for ``keydata`` and
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -155,12 +155,12 @@ Autocrypt-capable MUAs that implement :ref:`Gossip <key-gossip>` should
 also associate the following additional attributes with
 ``peers[addr]``:
 
-* ``gossip_timestamp``: the UTC timestamp of the most recent effective
-  date of all messages containing a valid ``Autocrypt-Gossip`` header
-  about the peer.
-* ``gossip_key``: the value of the ``keydata`` attribute derived from
-  the most recent message containing a valid ``Autocrypt-Gossip``
-  header about the peer.
+* ``gossip_timestamp``: the UTC timestamp of the most recent
+  ("youngest") ``origin-timestamp`` attribute in all messages
+  containing a valid ``Autocrypt-Gossip`` header for this peer.
+
+* ``gossip_key``: the value of the ``keydata`` attribute from
+  the youngest ``Autocrypt-Gossip`` header for the peer.
 
 How this information is managed and used is discussed in :ref:`peer-management`.
 
@@ -513,11 +513,13 @@ If both ``public_key`` and ``gossip_key`` are ``null``, then set
 Otherwise, we derive the recommendation using a two-phase algorithm.
 The first phase computes the ``preliminary-recommendation``.
 
+.. _`preliminary-recommendation`:
+
 Preliminary Recommendation
 __________________________
 
 If either ``public_key`` is ``null``, or ``autocrypt_timestamp`` is
-more than 35 days older than ``gossip_key_timestamp``, set
+more than 35 days older than ``gossip_timestamp``, set
 ``target-keys[to-addr]`` to ``gossip_key`` and set
 ``preliminary-recommendation`` to ``discourage`` and skip to the
 :ref:`final-recommendation-phase`.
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -637,29 +639,44 @@ Key Gossip
...
+
+A MUA MUST gossip the same public key for a recipient that
+it uses to encrypt to that recipient. The values for ``keydata`` and
+``origin-timestamp`` are thus computed mirroring the key selection
+logic used in `preliminary-recommendation`_:
+
+- If ``peer-state[gossip-addr]`` contains ``null`` in its
+  ``public_key`` attribute, or its ``autocrypt_timestamp`` is
+  more than 35 days older than its ``gossip_timestamp``, set
+  ``keydata`` to ``gossip_key`` and ``origin_timestamp``
```
Valodim: origin-timestamp

doc/level1.rst
```diff
@@ -637,29 +639,45 @@ Key Gossip
 ++++++++++
 
 It is a common use case to send an encrypted mail to a group of
-recipients. To ensure that these recipients can encrypt messages when
-replying to that same group, the keys of all recipients can be
-included in the encrypted payload. This does not include BCC
+recipients. To help these recipients to send encrypted replies
```
Valodim: what's wrong with "ensure"? it's really what is happening, isn't it?

dkg: there are some minor exceptions (e.g. replying to messages after a key has expired, even though the original message was sent before expiry), but they're narrow enough that i think "ensure" is acceptable here.

doc/level1.rst
```diff
@@ -637,29 +639,45 @@ Key Gossip
...
-See example in :ref:`autocrypt-gossip-example`
+See example in :ref:`autocrypt-gossip-example`.
+
+.. _`gossip-injection`:
 
 Key Gossip Injection in Outbound Mail
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-An Autocrypt MUA MAY include ``Autocrypt-Gossip`` headers in messages
+A MUA MAY include ``Autocrypt-Gossip`` headers in messages
```
dkg: should be "An MUA" -- but why remove Autocrypt here?  we're talking about Autocrypt-enabled MUAs, right?

doc/level1.rst
```diff
@@ -637,29 +639,45 @@ Key Gossip
...
 recipients, which by definition must not be revealed to other
 recipients.
 
-The ``Autocrypt-Gossip`` header has the format as the ``Autocrypt``
+The ``Autocrypt-Gossip`` header has the same format as the ``Autocrypt``
 header (see `autocryptheaderformat`_). Its ``addr`` attribute
 indicates the recipient address this header is valid for as usual, but
 may relate to any recipient in the ``To`` or ``Cc`` header.
-See example in :ref:`autocrypt-gossip-example`
+See example in :ref:`autocrypt-gossip-example`.
```
dkg: If this changeset is introducing a new attribute that is specific to `Autocrypt-Gossip` shouldn't it be introduced here?

doc/level1.rst
```diff
@@ -637,29 +639,45 @@ Key Gossip
...
+logic used in `preliminary-recommendation`_:
+
+- If ``peer-state[gossip-addr]`` contains ``null`` in its
+  ``public_key`` attribute, or its ``autocrypt_timestamp`` is
+  more than 35 days older than its ``gossip_timestamp``, set
+  ``keydata`` to ``gossip_key`` and ``origin_timestamp``
+  to ``gossip_timestamp``.
+
+- Otherwise, set ``keydata`` to ``public_key`` and
+  set ``gossip_timestamp`` to ``autocrypt_timestamp``.
```
dkg: This puts Recommendation logic in the Gossip Injection section.  I think if we're going to modify or augment the recommendation, it should happen in the section that talks about the recommendation.

maybe the output of the recommendation algorithm needs to be augmented?

doc/level1.rst
```diff
@@ -669,24 +687,28 @@ Updating Autocrypt Peer State from Key Gossip
...
+and only if all of the following conditions are true:
 
-2. If ``peers[gossip-addr].gossip_timestamp`` is more recent than the
-   message's effective date, then the update process terminates.
+- ``gossip-addr`` matches a recipient in the mail's ``To`` or ``Cc``
+  header,
 
-3. Set ``peers[gossip-addr].gossip_timestamp`` to the message's
-   effective date.
+- ``origin-timestamp`` is less recent that the current date,
```
dkg: should be "less recent than", or perhaps just "not in the future" -- or should it be the message's effective date?

doc/level1.rst
dkg: This leaks the time of last-received mail from this peer to everyone i'm mailing.

So if Alice mails Bob and Carol, Carol can see that Bob was writing Alice an e-mail at 4am?

Or, if we fuzz the resolution of the timestamp as @hpk42 suggested elsewhere, Carol can see that Bob wrote to Alice within the last week?

Why is this an acceptable tradeoff?

dkg:
I agree with @Valodim that the part about "SHOULD gossip the same keys used to encrypt" ought to be a separate PR, so we can discuss this proposed substantive change independently.

hpk42:
split out the gossip keydata/encryption consistency bit to #247

azul:
This discussion leaves me wondering if we really want to specify the logic to be used on conflicting gossip and peers keys. Let's keep in mind that this is a cornercase and hopefully a transitional state and there is no clear way to pick the right key.

I think we originally intended this specification to lead to consistent behaviour. However we are not even accomplishing this. Even if we add `origin-timestamp` - the behaviour depends on the difference to last-seen - so one peer might use the gossip key while another one uses the peers key because they saw a mail more recently.

My understanding is that this might even happen within the same threat. At least here it looks like `origin-timestamp` would lead to some eventual consistency.

What would happen if we stripped the key selection logic in case of conflicting keys from the spec and allowed the MUAs to come up with their own heuristic? There's a lot of things that a MUA might consider that we have not even included yet such as time a key was issued or signatures between the conflicting keys. My understanding is that there is even an OpenPGP way to signal that key A replaces key B.

PR #251, opened 2017-11-13
never prefer a gossip key over a directly received one. Fixes #248.

### LINKED ISSUE #248: Gossip key can override more recent "direct" key, potentially causing more unreadable mails
Suppose the follow scenario:

- T-42 days: ``Carol`` sends Alice a mail with an Autocrypt header, Alice responds.  

- T-41 days: Carol loses her device and reinstalls her mail app, resulting in a new Autocrypt key for future mails

- T-40 days: Carol writes bob a mail "damn i lost my device, " and otherwise they chat to tentatively agree for a joint date with Alice at time T. 

- T-30 days: Bob writes Alice a mail about jointly meeting at time T, she responds and says she will fix the time and location and send a mail around. 

- T-1: Alice writes Bob and Carol a mail about meeting location and precise T time for the three of them. Bob can read Alice's mail and group-replies with an attachment he wants to discuss and asks for feedback.  With the current logic, Bob will now use Carol's  stale key because Alice just gossiped it and his own state for Carol's key is from T-40 days, i.e. older than 35 days.   Alice and Bob exchange group-replies in preparation for the meeting.  Bob didn't hear back anything from Carol so also sends her a direct message. 

- T: Carol opens her mail client and can not read Alice's or bob's group mails and can not read Bob's direct mail either. She has no idea what is going on.  Didn't she successfully chat with Bob at T-40 days? She  group-mails back but answering takes a while because Alice/Bob are travelling ...  They eventually manage to meet and jointly decide to disable Autocrypt because it's causing too much trouble. 

There are variants on this.  Effectively i think that it's usability-wise dangerous to prefer a gossip key over a directly received one by taking the message date when a key was gossiped as significant. In the above situation, if Bob would have used his T-40 key from Carol, Carol would be able to read his group-reply to Alice's invitation at least and also the direct message from Bob. And if Alice would take the Carol-key from Bob during their T-1 conversation, Carol would only ever have seen one unreadable mail (the original one from Alice) instead of many. 

The preliminary recommendation 35 day logic was only introduced very recently (#209). We have no implementations for it so i think we should be careful in considering this.

hpk42:
It fixes #248 in the sense that Carol will be able to read half the mails (either the ones from Alice, or the ones from Bob) whereas with the current logic and the #248 scenario Carol can not read mails from neither bob or Alice. IOW,   50% unreadable mail is better than a 100%. Also, there isn't sufficient information for Bob to conclude that Alice's gossiped keys for carol are better than his own.  The date when a key was gossiped does not indicate much.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -525,8 +525,7 @@ The first phase computes the ``preliminary-recommendation``.
 Preliminary Recommendation
 __________________________
 
-If either ``public_key`` is ``null``, or ``autocrypt_timestamp`` is
-more than 35 days older than ``gossip_key_timestamp``, set
+If ``public_key`` is ``null``, then set
 ``target-keys[to-addr]`` to ``gossip_key`` and set
 ``preliminary-recommendation`` to ``discourage`` and skip to the
 :ref:`final-recommendation-phase`.


dkg:
While this change does mean that some threads between 3 parties that used to be 100% unreadable to one party will now be 50% unreadable, it also means that some threads that used to be 100% readable will now be only 50% readable.  right?

I just want to make sure we're ok with that.  It seems like if the goal is predictability, it'd be really nice to have all parties in a group conversation using the same keys for everyone.

hpk42:
yes, right.  Either Alice or Bob has a stale key and with this PR this will result in Carol being able to read either one or the other encrypted message, not both.  Without the PR, Carol will be able to read both in one case, and none in the other.  Having her MUA decrypt at leat one message, means she can group-reply encrypted, thus removing Alice/Bob's stale key.

Valodim:
I still find it odd that we have a concept of staleness, but don't apply it here. Both scenarios you mentioned are still possible even with stale keys, they just depend on whether it's before or after 35 days. We saw a mail recently and get a conflicting gossip key? It's very likely we have the better key. Otherwise? We don't know - just align with what the original sender of the message did, keep the thread consistent. Felt about right to me.

Anyways I can live with this. We will very likely have to fine tune when we get more experience with this no matter what we decide now.

azul:
I prefer this for Level 1 over #254 because it reduces complexity rather than increasing it.

Most of the discussion here is about the reply to encrypted situation it seems. However these are the recommendation for any mail going to a peer that has a conflicting gossip key.

@Valodim I see your point about stale keys. I think the context is different though. If we have a stale key the prelimary recommendation is discourage. That's a defensive meassure against unreadable mail. If we replace the key with the gossip key at the same time that adds another uncertainty to the mix. Now the key used between two users does not only depend on their messages exchanged but even on the gossip of other people.

So i suggest to merge this PR and thus always prefer keys we received directly over gossip keys in direct conversation.

To solve the issue with group replies and inconsistent keys I propose to encrypt to both keys... the latest gossip key in a conversation and the locally stored key. But that should be in the section about gossip and limited to the threat involved if possible.

PR #252, opened 2017-11-14
UI should not demand Passphrases - fixes #233

### LINKED ISSUE #223: Provide guidance on dealing with message signatures
When displaying an encrypted message, it's not currently clear how we expect autocrypt-capable MUAs to validate the signature.  Even given an attempt to validate the signature, it's not clear how (or whether) we expect the MUA to render it.

Options for Validation
================

 * never validate signatures at all (this is the easiest, and perhaps we should explicitly mention it for the Level 1 spec)
 * validate the signature of a message based on the public key found in the message's on ``Autocrypt`` header during message ingestion, and save the outcome of that validation alongside the message.  Perhaps also tabulate that information in the ``peers`` table somehow. (gathering statistics over time for local a reputation cache?)
 * validate the signature of a message during message rendering.  This presents (at least) two problems: 
    * which key do you validate it against?  the one currently in the ``peers`` table?  or the one in the message itself?
     * what if the message is rendered significantly later than when you ingested it (e.g.viewing archival messages)?  If there is any sort of signature or key expiration, it's conceivable that the signature would be invalid at rendering time.

Options for Rendering
================

* Never render the validation status of messages to the user at all (this is the simplest option, and might warrant being mentioned in the Level 1 spec).
* When the user asks for a message to be rendered, if the message was validated when it was ingested, display a simple symbol/icon (e.g.,✔) to indicate that it was valid at ingestion time.
* A user who fiddles with that icon might be able to expose any statistical information about that signing key, or the history that they have with it.

Open Questions
============
If a signature is missing, or does not validate, should the UI display anything?  [Recent discussions on the enigmail mailing list](https://admin.hostpoint.ch/pipermail/enigmail-users_enigmail.net/2017-November/004683.html) suggest just being silent in that situation.

Do we care about rendering signature validation for cleartext mail?  the same user may receive both encrypted and cleartext mails from the same peer.

Is there a way that a peer could announce that they *want* unsigned e-mails that purport to be from their address to show a warning?  This is sort of a rough equivalent of [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security).  Perhaps there could be a non-critical Autocrypt attribute that indicates this preference?  This opens the door to some kind of denial of service attack, though.  And it's not clear how it interacts with key rollover or other transition scenarios.

### LINKED ISSUE #233: change from "The MUA SHOULD NOT protect the secret key with a password." to UX
We don't actually care about whether the key is protected with a password.  we care about whether the user has to enter the password to read or send mail.

If the key is protected with a password that is unlocked upon session login, or upon "opening my mail app", i don't think that would be objectionable.

So this requirement should be phrased in terms of UX specifically -- what kind of interaction should be required (or forbidden) and when?  but not in terms of "is a password used".  The sentence before the objectionable one is actually pretty good:

> The secret key material should be protected from access by other applications or co-tenants of the device
> at least as well as the passwords the MUA retains for the user’s IMAP or SMTP accounts.

lefherz:
closes #233. 

the wording is not perfect of course, feel free to change it.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -740,8 +740,9 @@ encrypted outgoing messages MUST be signed, so if the user has to
 enter the password for both reading and sending mail, this introduces
 too much friction to become part of a routine daily workflow.
 
-Protection of the user's keys at rest and other files is achieved more
-easily and securely with full-disk encryption.
+Protection of the user's keys (and other sensitive data) at rest is
+achieved more easily and securely with filesystem-based encryption and
+other forms of access control.
 
 
 .. _multiaccounts:
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -736,9 +736,9 @@ passwords the MUA retains for the user's IMAP or SMTP accounts.
 The MUA MAY protect the secret key (and other sensitive data it has
 access to) with a password, but it SHOULD NOT require the user to
 enter the password each time they send or receive a mail. Since all
-encrypted outgoing messages MUST be signed, it introduces too much
-friction to become part of a routine daily workflow, if the user has
-to enter the password for both reading and sending mail.
+encrypted outgoing messages MUST be signed, it could happen that the
+user has to enter the password for both reading and sending mail. This
+introduces too much friction to become part of a routine daily workflow.
 
 Note that password protection of the secret key carries with it a risk
 that the user might forget their password, which might result in
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -726,12 +726,12 @@ keys instead.  The MUA MUST be capable of assembling these keys into
 an OpenPGP certificate (:rfc:`RFC 4880 "Transferable Public
 Key"<4880#section-11.1>`) that indicates these capabilities.
 
+Secret key protection at rest
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 The secret key material should be protected from access by other
 applications or co-tenants of the device at least as well as the
 passwords the MUA retains for the user's IMAP or SMTP accounts.
-
-Secret key protection at rest
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The MUA MAY protect the secret key with a password, but it SHOULD NOT
 require the user to enter it each time they send or receive a mail.
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -740,6 +740,16 @@ encrypted outgoing messages MUST be signed, so if the user has to
 enter the password for both reading and sending mail, this introduces
 too much friction to become part of a routine daily workflow.
 
+Note that password protection of the secret key carries with it a risk
+that the user might forget their password, which might result in
+catastrophic data loss.  Unlike IMAP or SMTP credentials (which can be
+reset by the server operator given some sort of out-of-band
+confirmation), there is no recovery workflow possible for the loss of
+a password protecting a secret key.  An MUA that chooses to offer
+password protection of the secret key (or other sensitive data) SHOULD
+support usable and secure backup/recovery workflows for the protected
+material.
+
 Protection of the user's keys (and other sensitive data) at rest is
 achieved more easily and securely with filesystem-based encryption and
 other forms of access control.
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -733,7 +733,7 @@ passwords the MUA retains for the user's IMAP or SMTP accounts.
 Secret key protection at rest
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The MUA MAY protect the secret key with a password, But it SHOULD NOT
+The MUA MAY protect the secret key with a password, but it SHOULD NOT
 require the user to enter it each time they send or receive a mail.
 All encrypted outgoing messages MUST be signed, so if the user has to
 enter the password for both reading and sending mail, this introduces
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -733,10 +733,12 @@ passwords the MUA retains for the user's IMAP or SMTP accounts.
 Secret key protection at rest
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The MUA SHOULD NOT protect the secret key with a password. All
-encrypted outgoing messages MUST be signed, which would require the
-user to enter the password for both reading and sending mail. This
-introduces too much friction to become part of a routine daily workflow.
+The MUA MAY protect the secret key with a password, But it SHOULD NOT
```
hpk42: But-> but

hpk42:
I guess @dkg rather wanted to have a note in the UI section but he can speak for himself. 

I am actually not much in favor of this change, but i am ok with it. Reason being that a passphrase can be forgotten if seldomly used which is a real-world usability problem.

lefherz:
Well, with the current text we don't want to force MUAs to forbid passphrases, but discourage them from enabling them.

But do we maybe want them to discourage passphrases?

Then only those who want to import their already used PGP keys into autocrypt would use this feature, and they tend to know their passphrases pretty well.

When we ask MUAs to discourage passphrases and recommend full-disk encryption, we enable experienced users to be as secure as they want. But still most users would go without passphrases and had the advantages of seamless encryption. This fits well into the UI section.

Apart from that, we should discourage MUAs from storing the keys in a way that it requires entering the passphrase on every message. This is not really a UI question and can stay here, I'd say.

dkg:
I like this series, and am writing a couple more patches that i think will take @hpk42's concerns into account here.

I think this is the right place to put the text, since we're giving a specific UI prescription to avoid a certain workflow.  If we were to suggest or require some other specifc workflow, then i'd want to put that suggestion in the UI section.

doc/level1.rst
```diff
@@ -726,19 +726,33 @@ keys instead.  The MUA MUST be capable of assembling these keys into
...
+too much friction to become part of a routine daily workflow.
+
+Note that password protection of the secret key carries with it a risk
+that the user might forget their password, which might result in
+catastrophic data loss.  Unlike IMAP or SMTP credentials (which can be
+reset by the server operator given some sort of out-of-band
+confirmation), there is no recovery workflow possible for the loss of
+a password protecting a secret key.  An MUA that chooses to offer
+password protection of the secret key (or other sensitive data) SHOULD
+support usable and secure backup/recovery workflows for the protected
```
b3yond: Backup/recovery workflows for lost passphrases? How is this possible without making passphrases totally useless?

dkg: Maybe that's an exercise for the reader?  :)  Note that the backup/recovery workflows aren't for the password itself, but for the material protected by the password.

Here are some examples:

 * secret-sharing split among trusted contacts
 * [the `paperkey` package](http://www.jabberwocky.com/software/paperkey/)
 * a hand-written Setup Code, stored offline in a safe, combined with the associated Autocrypt Setup Message

I think describing concrete examples of backup/recovery is out of scope for Level 1, but i do think we should mention it for implementers so that they don't accidentally burn their users.

doc/level1.rst
```diff
@@ -726,19 +726,33 @@ keys instead.  The MUA MUST be capable of assembling these keys into
...
-The MUA SHOULD NOT protect the secret key with a password. All
-encrypted outgoing messages MUST be signed, which would require the
-user to enter the password for both reading and sending mail. This
-introduces too much friction to become part of a routine daily workflow.
-Protection of the user's keys at rest and other files is achieved more
-easily and securely with full-disk encryption.
+The MUA MAY protect the secret key (and other sensitive data it has
+access to) with a password, but it SHOULD NOT require the user to
+enter the password each time they send or receive a mail.  All
+encrypted outgoing messages MUST be signed, so if the user has to
```
Valodim: This "MUST" kind of tripped me up more than once reading this sentence at different times. Can we change the sentence structure to a "Since", to indicate earlier on that this "MUST" is just leading up to something, rather than specifying that something MUST be one way or another? (arguably, it should actually be a "must" because of this, but that will trip other people up so better don't change that)

doc/level1.rst
b3yond: then I have no objections :)

doc/level1.rst
```diff
@@ -726,19 +728,34 @@ keys instead.  The MUA MUST be capable of assembling these keys into
...
-user to enter the password for both reading and sending mail. This
-introduces too much friction to become part of a routine daily workflow.
-Protection of the user's keys at rest and other files is achieved more
-easily and securely with full-disk encryption.
+The MUA MAY protect the secret key (and other sensitive data it has
+access to) with a password, but it SHOULD NOT require the user to
+enter the password each time they send or receive a mail. Since
+Autocrypt-enabled MUAs :ref:`sign all encrypted outgoing
+messages<mail-encryption>`, it could happen that the user has to enter
+the password for both reading and sending mail. This introduces too
```
Valodim: "has to enter the password very often, both for reading and sending mail"?

dkg: this kind of change is unobjectionable -- please add it as a patch on top of the existing series.

dkg: I've pushed it myself now.

PR #254, opened 2017-11-14
another try to fix #248 

### LINKED ISSUE #248: Gossip key can override more recent "direct" key, potentially causing more unreadable mails

hpk42:
This PR aims to cause more readable mails for Carol compared to the current spec state.  It should be at least as good as the current
spec state but gives parties receiving a gossip key a single bit of information: whether the gossiping party got the key directly or indirectly within the last ~35 days.

With this PR and the #248 scenario:

- Alice sends a Gossip-Header for Carol without the "fresh=true"
  setting in her group message to Bob & Carol

- Bob sends a mail using his own Carol key - Carol can read
  it because Bob has the correct key in the #248 scenario.

If instead Alice has a "fresher" key:

- Alice sends a Gossip-Header for Carol wit the "fresh=true"
  setting in her group message to Bob & Carol

- Bob will take Alice's key for Carol if he didn't
  receive a message from Carol in the 35 days before
  gossip_timestamp.

Effectively, the current spec state only works nicely if Alice has fresh keys, but fails the #248 case.  With this PR both cases work. 

And a note regarding "contact privacy": the gossiping party leaks the "fresh" bit of information (she might have gotten it via gossip, so not that much info). However, without this PR anyone who does not use the gossip key in their replies, leaks that information as well.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -168,6 +168,10 @@ also associate the following additional attributes with
 * ``gossip_key``: the value of the ``keydata`` attribute derived from
   the most recent message containing a valid ``Autocrypt-Gossip``
   header about the peer.
+* ``gossip_fresh``: A boolean which is true if the most
+  recent message containing a valid ``Autocrypt-Gossip`` header
+  also contained a ``fresh=yes`` setting.
+
 
 How this information is managed and used is discussed in :ref:`peer-management`.
 
@@ -525,8 +529,8 @@ The first phase computes the ``preliminary-recommendation``.
 Preliminary Recommendation
 __________________________
 
-If either ``public_key`` is ``null``, or ``autocrypt_timestamp`` is
-more than 35 days older than ``gossip_timestamp``, set
+If either ``public_key`` is ``null``, or ``gossip_fresh`` is True
+and ``autocrypt_timestamp`` is more than 35 days older than ``gossip_timestamp``, set
 ``target-keys[to-addr]`` to ``gossip_key`` and set
 ``preliminary-recommendation`` to ``discourage`` and skip to the
 :ref:`final-recommendation-phase`.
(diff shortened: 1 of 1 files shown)


hpk42:
sidenote: i am fine with getting the simpler #251 merged instead -- just don't like the current state which i think is not much cross-checked  (it btw has a non-existent ref to "gossip_key_timestamp").

hpk42:
as #251 got merged, i retract this one.

PR #257, opened 2017-11-15
update optional state to current level 1

azul:
I removed the OPTIONAL: inside the message processing because
the whole process is optional - but once you do it none of the
steps are optional anymore. If you leave them out the count will
be wrong.

I also removed references to ``autocrypt_peer_state[A]`` rather
than updating them to ``peers[P]`` to improve readability and
make this less coupled to the higher level data structures.

We do not have a reset state anymore - therefore we now compare
``counting_since`` to the ``autocrypt_timestamp``. This is
effectively a detection of resets because we update
``counting_since`` on resets.

### DIFF
--- a/doc/optional-state.rst
+++ b/doc/optional-state.rst
@@ -37,27 +37,24 @@ Managing additional state
 
 When processing a message from the peer:
 
- - OPTIONAL: If ``counting_since`` is unset, set it to the current time.
-   Otherwise, if ``message_date`` is greater than ``counting_since``:
+ - If ``counting_since`` is unset, set it to the current time.
+   Otherwise, if ``effective_date`` is greater than ``counting_since``:
 
-   - If ``pah`` is ``null``, increment ``count_no_ac``.
-   - If ``pah`` is not ``null`` increment ``count_have_ac``.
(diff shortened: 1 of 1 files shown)


azul:
On a closer look the behaviour here changed in case of a reset within the 35 days limit and another mail without an autocrypt header afterwards.

Previously the reset would not trigger the counters to be reset because it was in the 35 days limit. The second mail without an autocrypt header would also not reset the counters and update ``counting_since``. In the new version ``counting_since`` will be updated on the second autocrypt header that arrived after the 35 days.

We could have a better detection of what previously was a reset if we compare the last_seen and autocrypt_timestamp before processing the mail. If they were the same and now we analyse a mail without a ac header this is a reset. However this comparison has to happen BEFORE processing the mail. Which adds complexity to the overall algorithm. Maybe we could shift the whole optional processing to happen before the other.

Not sure what to think about that.
@BjarniRunar, can you explain the reasoning behind the algorithm a bit?

hpk42:
Looks good to me.  Merging this is definitely an improvement over the current state, despite the open questions.  @azul maybe you could put your suggestion/comment addressed to @BjarniRunar in an issue and we can close here?

PR #265, opened 2017-11-17
Clarify the Setup Message Content-Disposition type

### LINKED ISSUE #264: Autocrypt Setup Message: application/autocrypt-setup Content-Disposition
We don't say the Content-Disposition allowed in https://github.com/autocrypt/autocrypt/blob/master/doc/level1.rst#L836, can be both inline or attachment?. 
In the appendix example https://github.com/autocrypt/autocrypt/blob/master/doc/appendix/example-setup-message.eml, initially generated by Enigmail, it's attachment.
We also don't say whether html is allowed in the text above or below the ASCII-armored encrypted data, though probably this is less relevant, as that part is ignored while processing. We have html in the example.

juga0:
Closes #264

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -843,9 +843,12 @@ both programmatically and manually.
   allows implementations to optionally add another human-readable
   explanation.
 
-- The second mime part ``Content-Disposition`` SHOULD be ``attachment``
-  to facilitate the user to download it as a separate file. If it contains
-  HTML, the ``filename=`` MIME parameter should have the extension``.html``
+- The second MIME part's ``Content-Disposition`` SHOULD be
+  ``attachment`` to make it easy for the user to download it as a
+  separate file. If the second MIME part contains HTML, the
+  ``filename=`` MIME parameter should end with ``.html`` to make such
+  a file more likely to display as intended if the user tries to open
+  it.
 
 - The encrypted payload MUST begin with an ASCII-armored :rfc:`RFC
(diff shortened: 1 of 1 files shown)


hpk42:
looks good to me.

dkg:
this is a good improvement to the spec!  I did a bit of wordsmithing and added a brief justification for the ".html" business.  I think we should merge it.

doc/level1.rst
```diff
@@ -843,6 +843,13 @@ both programmatically and manually.
   allows implementations to optionally add another human-readable
   explanation.
 
+- The second MIME part's ``Content-Disposition`` SHOULD be
+  ``attachment`` to make it easy for the user to download it as a
+  separate file. If the second MIME part contains HTML, the
```
Valodim: I feel like this second sentence is distracting here. It opens up the whole deal about a "saveable as html" attachment type trickery, which is not trivial and will leave a reader wondering here with no further reference (e.g. search for "html" in the document) on what's going on.

This whole recommendation is sort of a side note (which is also why it's a  "should" rather than "SHOULD" I assume), I think I would either put it in a paragraph of its own that actually describes the idea, or leave it out for brevity. Given how speculative the idea is so far, I would go with the latter for 1.0.

The part about the `Content-Disposition` on the other hand is short and straightforward, it could be included in the previous paragraph right next to the specification of the same part's `Content-Type`.

If anyone +1s this I'll make a PR with this alternative suggestion.

dkg: I'd be happy to see an alternate PR.

PR #266, opened 2017-11-19
Eliminate plain text body, as the Email is multipart/encrypted

juga0:
There should not be a plain text string there.
Also, i'm not sure whether ``MIME-Version: 1.0`` should be in the description and payload parts, as it is in the ``multipart`` part.

### DIFF
--- a/doc/appendix/example-gossip.eml
+++ b/doc/appendix/example-gossip.eml
@@ -41,7 +41,6 @@ Content-Type: multipart/encrypted;
  protocol="application/pgp-encrypted";
  boundary="PLdq3hBodDceBdiavo4rbQeh0u8JfdUHL"
 
-This is an OpenPGP/MIME encrypted message (RFC 4880 and 3156)
 --PLdq3hBodDceBdiavo4rbQeh0u8JfdUHL
 Content-Type: application/pgp-encrypted
 Content-Description: PGP/MIME version identification


lefherz:
Why does this fail to build? o.0

juga0:
I also don't know why openssl complains about "-iv", when it doesn't in other commits [0] and .travis.yml is the same.
Closing this PR and reopening to force other build.

[0] https://travis-ci.org/autocrypt/autocrypt/builds/302961309#L823

juga0:
Still fails. @azul ideas?

azul:
It fails because your repo does not have the travis variables set that we use to decrypt the deployment key. Will figure out a way to make this more robust.

azul:
Trying to address this in #269

dkg:
the plain text string here is known as the MIME multipart "preamble" [the MIME content-types RFC](https://tools.ietf.org/html/rfc2046#page-21) explains:

> It is to be ignored, though it is a handy place for composition agents to include an explanatory note to non-MIME conformant readers.

It is common for MIME implementations to include such a preamble, so I think the example should include it.  I don't think we should merge this pull request.

Valodim:
Personally, I don't think preamble descriptions of multipart/encrypted are worth the bytes. They are seen by humans exceedingly rarely, even more rarely actually provide them with new information, and even in  those very rare cases much better information can be trivially found with the most naive google query ("multipart encrypted" yields the relevant rfcs for s/mime and openpgp in the first two results).

That said, they don't particularly hurt so I don't mind having them in the example either ¯\\_(ツ)\_/¯

More opinions? @azul @hpk42

dkg:
What if it said "this is the MIME preamble" instead?

Valodim:
That would reduce the usefulness even more, leaving a minor unrelated information for people that never saw a preamble in a multipart. I think I'd prefer the descriptive text over that.

There is also the thing about unnecessary metadata and fingerprinting clients etc...

dkg:
unnecessary metadata and fingerprinting clients are legitimate concerns.  perhaps we should actively campaign for message encryption toolkits to omit them when generating multipart/encrypted messages.

however, i think that question is orthogonal to what we expect to be in the example message, and to what we expect message decryption toolkits to be able to parse and deal with.  if we omit the preamble in the example, and someone implements a client that chokes on messages that have a preamble, they'll also choke on real-world messages.  I know @Valodim doesn't believe in Postel's law, but i do think we should encourage user agents to be able to cope with multipart-encrypted messages that have a preamble.

Valodim:
Telling clients how to correctly parse the preamble of a correct mime part is definitely out of scope for autocrypt proper. If we want to do something like that, a collection of references to build on, or a set of test messages would be a better way to go.

dkg:
The spec is not telling them how to parse it, nor should it.  fwiw, i don't want them to parse it, i want them to ignore it.  But the example message in the Autocrypt spec has a preamble because MIME allows it.

the Autocrypt spec also doesn't tell clients how to parse the Date: header, or how to structure a MIME message.  but the examples are structured MIME messages, and contain Date: headers that we expect clients to parse.

Valodim:
A mime structure and Date header are arguably part of a complete example email. By comparison, having a preamble in our human-readable spec doesn't seem very useful to me if we *don't* want clients to actually include one. I would merge this PR, but again, I don't have particularly strong feelings about this. Feel free to merge or close this PR, or flip a coin for it :)

Valodim:
In the interest of progress, let me rephrase: Feel free to close this PR, otherwise I will merge it tomorrow evening.

PR #268, opened 2017-11-26
New server

fschulze:
This adds the new server as deployment target. Once this is merged, I can update DNS. After a few days we can remove the old server.


fschulze:
Actually someone else needs to update DNS, it's not in the places I have access to.

PR #269, opened 2017-11-26
deploy: only attempt if env vars are set accordingly

azul:


azul:
Looks like #268 also addresses the underlying problem. Closing this for now.

Issue #267, opened 2017-11-24
how to transfer multiple secret keys in setup message? 


hpk42:
Brought up by a review from Benny from the L1.0.0rc1-Nov22 built: what is the current idea on how to transfer multiple secret keys in the setup message?  

use case: If a MUA regenerates a secret key (i.e. sends out a new key) it still wants to retain the former secret key in order to be able to decrypt messages encrypted to it.  This is neccessary as there is no way to push out key updates to peers other than regular e-mail messages.  If such a MUA with two secret keys wants to transfer both secret keys to another autocrypt-capable MUA, how would it construct the message?

Could it add a third ``application/autocrypt-setup`` mime part to the ``multipart/mixed`` [message structure](https://autocrypt.org/en/latest/level1.html#message-structure)?  and could this be specified in a 1.1 backward-compatible release of the spec?  

it seems that the current wording forbids having multiple secret keys in the ``application/autocrypt-setup`` part as it "MUST" ignore anything before and after the single secret key. 

i think this issue needs to be clarified before 1.0 final (not neccessarily resulting in a spec change, but maybe into e.g. a FAQ entry).

freron:
A few extra thoughts on the subject:

It seems to me that it would be fine to allow the `multipart/mixed` part to have multiple `application/autocrypt-setup` subparts. The first one should be the one used for new messages -- as it is now. Maybe specify that the first one MUST be processed and the others SHOULD be processed.

Also, the passphrase must be the same for all of the keys in order to avoid entering multiple setup codes. An expert (not me) should check if this somehow weakens the security.

Given the above then it could also be emphasized that the MUA should only process the most recent “Autocrypt Setup Message” found for a given identity (if the MUA supports automatically looking for these).

dkg:
Just a note of clarification -- the current setup message is *intended* to transport two secret key mathematical objects in the `application/autocrypt-setup` part -- the secret key for the primary key, and the secret key for the subkey.  I know that this is all part of a single "OpenPGP Transferable Secret Key", and i think that this issue is asking about multiple "transferable secret keys" instead, though.

The current spec does not contemplate retention of any previously-held secret keys for an account.  I agree that this is something a sensible MUA might want to do, but we don't spec it out at all at the moment (e.g. there is no `accounts[A].historical_secret_keys` field), which makes it tough to specify how a user agent might handle multiple keys in a setup message.

[...]

PR #275, opened 2017-12-12
make sure parsing stops after the first ascii armor ending delimiter

Valodim:
This is a slight improvement for handling autocrypt setup messages. See #267 for a discussion


r10s:
Should we mention explicitly that stuff before the first ascii armor starting delimiter must also be ignored?
Just in case we decide to add sth. before later. Or do we switch to v2 then?

dkg:
Right now it says that the cleartext MUST start with the ASCII-armor starting delimiter.  I think that's a perfectly fine constraint, and i wouldn't want to relax it.

PR #279, opened 2017-12-14
minor: fix caps on landing page

azul:


hpk42:
+1 but i'd like a native english speaker to merge as i am always unsure about native expectations ...

PR #281, opened 2017-12-14
Correct the Autocrypt Setup Message's encrypted content

dkg:
OpenPGP's Symmetrically-Encrypted Data Packet has no integrity
protection, and it should never be used.

Furthermore, the setup message should use an SKESK, both to provide
some level of PBKDF2-based key stretching, and to ensure that the
actual symmetric key used is a full-fledged AES-128 key.

This also matches what GnuPG will generate when running ``gpg
--symmetric``.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -838,8 +838,13 @@ both programmatically and manually.
 - The second mime part of the message MUST have Content-Type
   ``application/autocrypt-setup``, and SHOULD have Content-Disposition
   of ``attachment``. Its content consists of the user's ASCII-armored
-  secret key, encrypted in an ASCII-armored :rfc:`RFC 4880
-  Symmetrically Encrypted Data Packet<4880#section-5.7>`
+  secret key, encrypted within an ASCII-armored OpenPGP
+  symmetrically-encrypted message.  Specifically, this means a block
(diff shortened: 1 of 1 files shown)


pbrunschwig:
fine with me

r10s:
looks good to me, in fact, this is what I use in delta.

doc/level1.rst
```diff
@@ -837,8 +837,13 @@ both programmatically and manually.
 - The second mime part of the message MUST have Content-Type
   ``application/autocrypt-setup``, and SHOULD have Content-Disposition
   of ``attachment``. Its content consists of the user's ASCII-armored
-  secret key, encrypted in an ASCII-armored :rfc:`RFC 4880
-  Symmetrically Encrypted Data Packet<4880#section-5.7>`
+  secret key, encrypted with a in an ASCII-armored OpenPGP
```
r10s: i'm no english expert, but some of "with a in an" looks superfluous to me

hpk42:
+1 for merging except for the syntax glitch that @r10s already pointed out.

doc/level1.rst
dkg: good catch, @r10s, thanks

Issue #277, opened 2017-12-12
Need guidance for what to do when receiving a setup message for an account that we already have a secret key for


dkg:
Some MUA will have a secret key for account `A` already, and then they'll see an Autocrypt Setup Message for `A` that contains a different key.

What should the MUA do in that situation?  we need clear guidance so that we don't have wildly different behaviors.

dkg:
We currently only recommend looking for ASMs when enabling Autocrypt on an account for the first time.

Some options:

 a) completely ignore ASMs for an account if you already have a key for it.  If the user disables and then re-enables Autocrypt, just keep the old key.

 b) if the user disables and then re-enables, do the scan at that point, and offer them a choice between the old key and the discovered key (or discovered keys, in the event that more than one was found?)

 c) when disabling Autocrypt, offer the user an opportunity to destroy their key, and access to all of the e-mail that might have been encrypted to it.  this completely resets their state, which would allow them to scan for a new key at that point.

 d) do not require the user to disable/enable Autocrypt to scan for keys -- instead, next to enable/disable, offer a choice like "look for a setup message from other Autocrypt e-mail programs"

 e) during regular mail scanning, if an ASM is encountered for an account that is already locally autocrypt-enabled, actively offer the user a choice between the setup in the new ASM, and the existing setup.


Any other suggestions?  does anyone have any strong feelings about what the Right Thing™ is?  what about the Right Thing™ for Level 1?  (a) is certainly the easiest to do, but maybe the most frustrating for users, who will need to set up an entirely new MUA profile if they've gotten into that state.  Might be ok for L1, though.

(c) is appealing to me for L1 because:

 * it's simple to implement
 * the "delete my state" feature seems like it might be useful anyway, for privacy reasons.  Being able to confidently delete sensitive material is a good thing.
 * it doesn't require too much extra UI, or complicated questions.

r10s:
I would got for e) as the Right Thing™ - if there is a "fresh incoming ASM", show it to the user and ask if he wants to import the key and warn if there is already an existing one. Disabling Autocrypt should not delete existing keys.

Reasons: 

- normally, the user has initiated the process on another device, I think, this is a strong clue the user wants to use the key.
- easy to implement - in fact, I think this is already the current state for Enigmail and Delta - every ASM is shown as a - more or less - normal incoming message with a "start setup" button - it's up to the user to click or not to click.  
- few additional UI elements (a single question)
[...]

PR #285, opened 2017-12-16
Do not overwrite existing secret keys before asking the user

r10s:
discussion in #277

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -1036,7 +1036,7 @@ encrypt.
 Account Preferences
 +++++++++++++++++++
 
-Level 1 MUAs MUST allow the user to disable Autocrypt completely for
+Level 1 MUAs SHOULD allow the user to disable Autocrypt completely for
 each account they control (that is, to set ``accounts[addr].enabled``
 to ``false``).  For level 1, we expect most MUAs to have Autocrypt
 disabled by default.


doc/level1.rst
```diff
@@ -812,7 +812,10 @@ can generate an Autocrypt Setup Message, and send it to itself.  A
 not-yet-configured Autocrypt MUA (a new MUA in a multi-device
 case, or recovering from device failure or loss) can import the
 Autocrypt Setup Message and recover the ability to read existing
-messages.
+messages. If the Autocrypt MUA is already configured (MUA has secret key
+material in ``accounts[addr].secret_key``) then the UI MUST display a
+warning to the user and ask them to confirm that importing the Autocrypt
+Setup Message will overwrite the existing key.
```
r10s: maybe add "... and the user may loose access to all mails encrypted to this key."

hpk42: so there is then no way to import the key but retain the old?  I know we don't want to specify historic key state in the L1 spec but i'd rather give the advise that the MUA should keep the old key around for decryption purposes (not specified how) to avoid unreadable mail.  For Delta it's a little less of an issue because it keeps its internal device state independently from the pgp encryption.  for enigmail/k9 that's not the case i think. 

@Valodim or @pbrunschwig Would you for enigmail or K9 really want to throw away ("overwrite") a key that has been used and send out?

r10s: for Delta Chat, in fact, I plan to keep the old keys, mainly for being able to decrypt new incoming messages still encrypted with the legacy key (already downloaded messages are stored decrypted).

however, I think if we add an general Autocrypt-advice to keep the legacy keys, we should also offer a plan about how to transfer legacy keys later.

pbrunschwig: I interpret "overwrite" as "overwrite the settings". In other words, I would change the configuration to use the new key. But Enigmail would certainly keep the old key as well.

dkg: This change seems like it's in an odd place in the spec.  The paragraph we're adding to is about the overall utility of the setup message, and the places where it becomes relevant -- when to generate the setup message, when to look for it, and what can be done when processing it.  That's already a lot!

But this change introduces still more: it introduces some additional UI guidance for setup message processing in a context in which we have not yet contemplated even looking for the setup message.

That is, up until this addition, the already-configured Autocrypt MUA apparently has no reason to look at the setup mesage.  Now, we're both giving them a reason, and including some minor guidance of what to do.  

I'd expect this sort of guidance to show up in the `Setup Message Import` section instead, but even that section doesn't yet contemplate the already-configured use case.

r10s:
I would suggest to close this PR in favour to #287

dkg:
ok, closing.  I hope you'll consider writing up an edit for post-level-1 clients that documents better practices for historical keys :)

PR #286, opened 2017-12-18
point media inquiries to "press" e-mail alias.

hpk42:
added an email possibility to contact at the autocrypt domain, which currently comprises dkg, valodim, bjoern and me -- please contact me personally if you want to get added. The alias is hosted on mail.autocrypt.org which i'd like to shift to the autocrypt.org vm proper (along with the bot).

### DIFF
--- a/doc/contact.rst
+++ b/doc/contact.rst
@@ -16,9 +16,9 @@ If you want to talk with us you may:
 - join chats at **#autocrypt on freenode** or through `matrix mirror
   <https://riot.im/app/#/room/#autocrypt:matrix.org>`_.
 
-- if you are writing articles or want to more privately talk to project
-  members please mail contact at this website domain. Any technical
-  questions should go to the above public channels instead, though.
+- if you want to more privately talk to project members please mail
+  contact at autocrypt.org. Any technical questions should go to
+  the above public channels instead, though.
 
 .. _`Autocrypt mailing list`: https://lists.mayfirst.org/mailman/listinfo/autocrypt
 
--- a/doc/contact.rst
+++ b/doc/contact.rst
@@ -16,6 +16,9 @@ If you want to talk with us you may:
 - join chats at **#autocrypt on freenode** or through `matrix mirror
   <https://riot.im/app/#/room/#autocrypt:matrix.org>`_.
 
+- if you are writing articles or want to more privately talk to project
+  members please mail contact at this website domain. Any technical
+  questions should go to the above public channels instead, though.
 
 .. _`Autocrypt mailing list`: https://lists.mayfirst.org/mailman/listinfo/autocrypt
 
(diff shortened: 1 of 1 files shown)


dkg:
Why not spell out the e-mail address explicitly?  saying "contact at this domain" without even setting off "contact" as a literal string makes it very difficult to tell that this is an e-mail address.

Do we want a  security-reporting alias as well, in case there are security concerns raised with the spec?

how do we want to handle replies to messages sent to this alias? For example, should everyone who receives mail on this alias who wants to reply send the reply from themselves, but also cc-ing the alias?

do we need an Autocrypt key for this alias?

doc/contact.rst
```diff
@@ -16,6 +16,9 @@ If you want to talk with us you may:
 - join chats at **#autocrypt on freenode** or through `matrix mirror
   <https://riot.im/app/#/room/#autocrypt:matrix.org>`_.
 
+- if you want to more privately talk to project members please mail
```
dkg: This sounds odd.  if someone wants to talk privately to members of the project, and they know which member they want to talk to, they can contact that member directly (it's easy to find our contact information by reading the mailing list archives).

So we're trying to handle a situation where they don't know which project member they want to talk to specifically, and they can't or don't want to use the public list.  right?

Let's try to walk through the specific reasons that they want to talk privately.  depending on those reasons, they may have different needs.

 * security concerns -- these folks might prefer encrypted e-mail.  Can we give it to them?
 * difficulty subscribing to the mailing list -- this should be addressed from the mailing list itself, not here
 * technical questions, but discomfort with discussing them publicly -- we're saying below that this is an inappropriate use of this address
 * media inquiries -- perhaps this deserves its own rationale, or its own e-mail address?
 * any other reason?

hpk42: that would be "press@"

dkg:
The current revision looks nice and simple.  I like it.  I don't know why it's not a `mailto:` link, though.  Any reason for the minimalist obfuscation?

Do we have internal policy about press inquiries?  is the alias already set up? -- who does it go to?  are people who respond to it expected to Cc the press address as well, so that other people know whether someone has already responded or not?

dkg:
Looking at `mail.autocrypt.org:/etc/aliases`, it doesn't look like  the `press` alias is set up, and i don't  think i have permissions to maintain that VM, so i can't fix it.  I think the alias needs to be tested and there needs to be consensus on-list about what to do with messages that come in on the alias before we merge this PR.

hpk42:
note that @valodim had asked me on IRC to quickly add a "contact" alias with him on it.  I did it (also adding me and @dkg) and now I am just trying to make it a bit transparent that we have a special mailing address and am not prepared currently to organize a whole discussion around it, sorry.  I rather need help with organizing background things (like "consensus on-list") than being able to do yet more of that myself.  So i am tentatively closing this PR here.  i think it's worthwhile to open an issue, and will next time point requests as from @valodim to the issue tracker and ask him to organize consensus first.

PR #287, opened 2017-12-18
Document disabling and destroying secret key material

dkg:
We currently have no discussion about the details of disabling
Autocrypt, or of how to expose the user to the choice of destroying
secret key material.

This change introduces this necessary text, and also observes that it
is the simplest/shortest path to synchronizing secret keys.

As such, it implements option (c) as discussed in
https://github.com/autocrypt/autocrypt/issues/277, and resolves this
problem for level 1 MUAs in a minimalist way.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -948,7 +948,8 @@ Setup Message Import
 
 An Autocrypt-capable MUA SHOULD support the ability to find and import
 an Autocrypt Setup Message when the user has not yet configured
-Autocrypt.  An MUA could look for such a message in several ways,
+Autocrypt (that is, when ``accounts[addr].secret_key`` is unset).  An
+MUA in this state could look for such a message in several ways,
 including:
 
  * If the user decides to enable Autocrypt for an account, and
@@ -1036,7 +1037,7 @@ Account Preferences
 Level 1 MUAs MUST allow the user to disable Autocrypt completely for
 each account they control (that is, to set ``accounts[addr].enabled``
 to ``false``).  For level 1, we expect most MUAs to have Autocrypt
-disabled by default.
+disabled by default. See :ref:`enabling-disabling` for more details.
 
 .. _getting-started:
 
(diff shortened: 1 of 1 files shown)


r10s:
I think the PR is clear and fits well into the whole spec.

Some comments:

- MUAs that rely on Autocrypt by default, may not have an option to disable it (or the disable option is to uninstall the app).
Of course, they can add a disable/enable option, however, for the purpose of just using another key this seems to be overkill.
My mindset for Delta Chat when receiving an ASM for an already configured account would be to "ask user - disable ac - enable ac - process ASM" in one single step :)
- @pbrunschwig: at least currently, Enigmail also processes ASM with an already configured Autocrypt setup, for me, this is the logical flow, not sure if this will change  
- unless I have overlooked something, there is still now way to keep legacy keys (okay, my PR didn't have this as well), however, I think in practise MUAs will just do this (see eg. [Patrick's comment and mine above](https://github.com/autocrypt/autocrypt/pull/285#discussion_r157380692))

dkg:
I'm fine if MUAs decide to keep legacy keys.  hopefully some of the MUAs that do this will contribute their design to a level 2 spec, and we can settle on a common/standard way to do it :)

@r10s wrote:
> MUAs that rely on Autocrypt by default, may not have an option to disable it

I think (with or without this PR) that is already kind of frowned on by the spec, which says: 

> Level 1 MUAs MUST allow the user to disable Autocrypt completely for each account they control (that is, to set `accounts[addr].enabled` to false).

I think that might be a separate issue from this PR, though.

doc/level1.rst
```diff
@@ -1082,6 +1083,48 @@ Earlier choices are better than later ones.
...
+When disabling Autocrypt for an account, a Level 1 MUA MAY offer the
+user an opportunity to also destroy the secret key material for that
+account.  Since Autocrypt clients generally do not discuss secret keys
+with users, a MUA offering this choice should use a phrase like
+"destroy access to all encrypted messages", rather than referring to
+"keys" or "key material".
+
+A MUA that allows the user this opportunity SHOULD clearly indicate to
+the user that the destruction of this secret key material will leave
+them unable to read their previously-received encrypted messages as
```
hpk42: whether or not past messages can be read depends on how a MUA handles storage of incoming messages. Delta keeps decrypted messages in its local store whereas e.g. k9/thunderbird use IMAP and thus pgp-encrypted messages for their message store.  

i guess it might be better overall to talk unreadable messages in user-facing warnings aka something like "future received message may be unreadable if you destroy your secrets" .

doc/level1.rst
```diff
@@ -1082,6 +1083,48 @@ Earlier choices are better than later ones.
...
+appropriate action to take in some circumstances.
+
+If the user selects this option, the MUA MUST clear both
+``accounts[addr].secret_key`` and ``accounts[addr].public_key``.
+
+After this key destruction has happened, the MUA is in a configuration
+where it could restart a :ref:`search for an Autocrypt Setup
+Message<setup-message-import>`.  For Level 1 MUAs, this is the
+simplest path for synchronizing secret keys between two MUAs that have
+mistakenly both enabled Autocrypt independently.
```
hpk42: I understand that you'd like to provide a path for users to deal with "MUA received an ASM but already has a secret" by way of saying "first destroy autocrypt secret, then re-scan for ASM". This might be useful in some situations but hurting others.  I suggest to remove this paragraph and leave the question how to deal with "dual" autocrypt state to the other PR (it can reference destruction of key state as stated in your PR if it needs to, no need to prescribe how we deal with dual autocrypt state here).

doc/level1.rst
dkg: Sure, a MUA that does *not* destroy access to the messages should not indicate that they will be destroyed if they are not destroyed.  I'll push a new revision that takes this into account.

doc/level1.rst
dkg: OK, i've gone ahead and updated this branch to drop the controversial paragraph as well.  @hpk42 are you OK with merging this?

PR #290, opened 2017-12-19
website: put the actual mission first

dkg:
When making a mission statement, the first sentence should be a simple
and clear statement of our goals.  It should not start with an
explanation of why we haven't reached them yet, or of the historical
background.  Those things are fine to include, but the first sentence
after "Our Mission" should be the mission of the project.

### DIFF
--- a/doc/_templates/customindex.html
+++ b/doc/_templates/customindex.html
@@ -40,7 +40,14 @@
   <div>
     <h2>Our Mission</h2>
     <p>
-      End-to-end encrypted e-mail has been around for decades, but has failed to see wide adoption outside of specialist communities.  Autocrypt aims to provide encryption convenient enough for much wider adoption. Through strengthening the e-mail ecosystem step-by-step, we want to increase user autonomy and privacy for everyone.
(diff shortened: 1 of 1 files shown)


doc/_templates/customindex.html
```diff
@@ -40,7 +40,14 @@ <h2>Open Standard</h2>
   <div>
     <h2>Our Mission</h2>
     <p>
-      End-to-end encrypted e-mail has been around for decades, but has failed to see wide adoption outside of specialist communities.  Autocrypt aims to provide encryption convenient enough for much wider adoption. Through strengthening the e-mail ecosystem step-by-step, we want to increase user autonomy and privacy for everyone.
+      Autocrypt aims to provide end-to-end encrypted e-mail that is
+      convenient enough for wide adoption.
+    </p>
+    <p>
```
hpk42: I agree with the shuffling around but less with making it into three instead of two paras.  Would you mind removing the extra para and just have the second sentence next the first?

hpk42:
i added an alternative PR #291 which does a lot more than your changes

PR #294, opened 2017-12-20
Document a simple workflow for when two MUAs conflict on an account

dkg:
This is a followup to #287.

This documents choice (c) of #277 and provides a
minimalist path forward to clear that concern for Level 1.

With this merged, i think we can clear the level 1 milestone from #277.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -1122,6 +1122,12 @@ take in some circumstances.
 If the user selects this option, the MUA MUST clear both
 ``accounts[addr].secret_key`` and ``accounts[addr].public_key``.
 
+After this key destruction has happened, the MUA is in a configuration
+where it could restart a :ref:`search for an Autocrypt Setup
+Message<setup-message-import>`.  For Level 1 MUAs, this is the
+simplest path for synchronizing secret keys between two MUAs that have
+mistakenly both enabled Autocrypt independently.
+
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -1122,6 +1122,12 @@ take in some circumstances.
 If the user selects this option, the MUA MUST clear both
 ``accounts[addr].secret_key`` and ``accounts[addr].public_key``.
 
+After this key destruction has happened, the MUA is in a configuration
+where it could restart a :ref:`search for an Autocrypt Setup
+Message<setup-message-import>`.  For Level 1 MUAs, this is the
+simplest path for synchronizing secret keys between two MUAs that have
+mistakenly both enabled Autocrypt independently.
+
```
hpk42: i think this paragraph should rather go to "Setup Message Import" section and link to the "destory" section here. I am fine with merging it thus because the wording says "this is the simplest" path which indicates it's not the only possible solution without diving deeper (a 1.1 topic i think).

dkg:
I just opened #300 as an alternative to this, trying to take into account @hpk42's feedback above.

At most one of these two PRs should be merged.  Once one is merged, the other should be closed.

hpk42:
thanks @dkg, i merged the other one!

PR #296, opened 2017-12-20
refined the wording

lefherz:
On https://readable.io/text/, this PR improves the readability from D (not so good) to B (quite ok). It is still not perfect, and that we have to use words like "specification" (5 syllables!) isn't leaving much more room^^

Anyway, I tried to do my best, and not touch the meaning too much. But probably I missed some important nuance. Just let me know :)

Apart from that, I think the "Test it now" button would look much better above this text and centered. Maybe also a bit larger. I couldn't make it work with css though, feel free to add commits to this PR which deal with this.

### DIFF
--- a/doc/_templates/customindex.html
+++ b/doc/_templates/customindex.html
@@ -38,13 +38,22 @@
     </div>
   </div>
   <div>
-    <h2>What is Autocrypt? What is the status?</h2>
+    <h2>What is Autocrypt?</h2>
     <p>
-      Autocrypt is a specification which mail application programmers can implement
-      to offer their users convenient end-to-end encryption for e-mail. The first specification
-      was released in December 2017 and a few mail programs already have or are heading 
-      towards Autocrypt-capable releases in spring 2018.  Please join us in implementing
-      and refining "Level 1" and other upcoming specifications. For any further questions
+      Autocrypt is not an app, but there are autocrypt-enabled
+      mail-apps. Autocrypt is a specification for convenient
+      end-to-end-encryption of E-Mails. Mail application
+      programmers are currently implementing it.
+    </p>
+    <h2>Does it already work?</h2>
+    <p>
+      The first specification was released in December 2017. A
+      few mail programs are already Autocrypt-capable, some want
+      to reach autocrypt-capability in spring 2018.
+    </p>
+    <p>
+      Please join us in implementing and refining "Level 1" and
+      other upcoming specifications. For any further questions
       or feedback, see our <a href="{{ pathto('contact') }}">Contact Page</a>.
     </p>
   </div>
--- a/doc/_templates/customindex.html
+++ b/doc/_templates/customindex.html
@@ -40,20 +40,20 @@
   <div>
     <h2>What is Autocrypt?</h2>
     <p>
-      Autocrypt is a specification for convenient
-      end-to-end-encryption of E-Mails. Mail application
-      programmers are currently implementing it.
+      Autocrypt is a set of guidelines for convenient
+      end-to-end-encryption of e-mails. Developers are working on
+      bringing it to several different e-mail programs.
     </p>
     <h2>Does it already work?</h2>
     <p>
-      The first specification was released in December 2017. A
-      few mail programs are already Autocrypt-capable, some want
-      to reach autocrypt-capability in spring 2018.
+      The first specification, Level 1, was released in December 2017.
+      A few mail programs are already capable of Autocrypt, some want
+      to be able of Autocrypt in Spring 2018.
     </p>
     <p>
-      Please join us in implementing and refining "Level 1" and
-      other upcoming specifications. For any further questions
-      or feedback, see our <a href="{{ pathto('contact') }}">Contact Page</a>.
+      Please join us in implementing and refining Autocrypt Level 1
+      and other upcoming specifications. For any further questions or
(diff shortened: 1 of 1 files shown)


doc/_templates/customindex.html
```diff
@@ -38,13 +38,22 @@ <h2>Open Standard</h2>
...
-    <h2>What is Autocrypt? What is the status?</h2>
+    <h2>What is Autocrypt?</h2>
     <p>
-      Autocrypt is a specification which mail application programmers can implement
-      to offer their users convenient end-to-end encryption for e-mail. The first specification
-      was released in December 2017 and a few mail programs already have or are heading 
-      towards Autocrypt-capable releases in spring 2018.  Please join us in implementing
-      and refining "Level 1" and other upcoming specifications. For any further questions
+      Autocrypt is not an app, but there are autocrypt-enabled
+      mail-apps. Autocrypt is a specification for convenient
```
hpk42: I'd strike the first sentence (Autocrypt is not an app ...).  It's also not a tool, not a software packages (what some commenters on heise believed) etc.  let's just directly say it's a specification.  somehow even end-users will likely guess that they can't install a specification especially with the follow up sentences.

lefherz: It is polemic, yes. But I think we can afford some polemics in the first sentence - they are cleared up directly afterwards, but improve understandability a lot.

With this sentence, people will know, what they can expect and what not. It helps to sort the complicated stuff into the right drawers.

hpk42: i didn't say it's polemic. also what autocrypt-enabled mail apps are there? The ones under "Test it" are not single-click there, are they?  Also "mail app" makes sense to you, but i have talked to people who aren ot familiar with this term because they consider mail to be "gmail" or "riseup" things (i.e. web interface in our terms).  

so I tentatively insist that just starting with "autocrypt is a specification" is best thing to do here.

doc/_templates/customindex.html
```diff
@@ -38,13 +38,21 @@ <h2>Open Standard</h2>
...
-    <h2>What is Autocrypt? What is the status?</h2>
+    <h2>What is Autocrypt?</h2>
     <p>
-      Autocrypt is a specification which mail application programmers can implement
-      to offer their users convenient end-to-end encryption for e-mail. The first specification
-      was released in December 2017 and a few mail programs already have or are heading 
-      towards Autocrypt-capable releases in spring 2018.  Please join us in implementing
-      and refining "Level 1" and other upcoming specifications. For any further questions
+      Autocrypt is a specification for convenient
+      end-to-end-encryption of E-Mails. Mail application
```
dkg: Why is E-Mails capitalized here?

doc/_templates/customindex.html
```diff
@@ -38,13 +38,21 @@ <h2>Open Standard</h2>
...
+      programmers are currently implementing it.
```
dkg: "Mail application programmers" is a little bit odd.  how about "Developers are working on adding it to several different e-mail programs."?

doc/_templates/customindex.html
```diff
@@ -38,13 +38,21 @@ <h2>Open Standard</h2>
...
+    </p>
+    <h2>Does it already work?</h2>
+    <p>
+      The first specification was released in December 2017. A
+      few mail programs are already Autocrypt-capable, some want
+      to reach autocrypt-capability in spring 2018.
+    </p>
```
dkg: why is Autocrypt-capable capitalized, but autocrypt-capability not?  also, autocrypt-capability sounds odd to my en_US-speaking ears.

doc/_templates/customindex.html
```diff
@@ -38,13 +38,21 @@ <h2>Open Standard</h2>
...
+    <p>
+      Please join us in implementing and refining "Level 1" and
```
dkg: This should probably be "Autocrypt level 1"

doc/_templates/customindex.html
```diff
@@ -38,13 +38,21 @@ <h2>Open Standard</h2>
...
-      and refining "Level 1" and other upcoming specifications. For any further questions
+      Autocrypt is a specification for convenient
+      end-to-end-encryption of E-Mails. Mail application
+      programmers are currently implementing it.
+    </p>
+    <h2>Does it already work?</h2>
+    <p>
+      The first specification was released in December 2017. A
+      few mail programs are already Autocrypt-capable, some want
+      to reach autocrypt-capability in spring 2018.
```
dkg: [common English rules](http://grammar.yourdictionary.com/capitalization/capitalization-of-seasons.html) would have you capitalize "Spring 2018"

doc/_templates/customindex.html
```diff
@@ -38,13 +38,21 @@ <h2>Open Standard</h2>
...
   <div>
-    <h2>What is Autocrypt? What is the status?</h2>
+    <h2>What is Autocrypt?</h2>
     <p>
-      Autocrypt is a specification which mail application programmers can implement
-      to offer their users convenient end-to-end encryption for e-mail. The first specification
-      was released in December 2017 and a few mail programs already have or are heading 
-      towards Autocrypt-capable releases in spring 2018.  Please join us in implementing
-      and refining "Level 1" and other upcoming specifications. For any further questions
+      Autocrypt is a specification for convenient
```
dkg: Is "specification" too much of a term of art?  what about "set of guidelines" instead?

Issue #295, opened 2017-12-20
make non-master websites render differently


dkg:
When rendering the web site from a non-master branch, it should be visually distinct from the master branch -- see the discussion in #292 .

I'm going to try to do this with sphinx.

fschulze:
This might help: https://docs.travis-ci.com/user/environment-variables#Default-Environment-Variables

You could check for ```TRAVIS_REPO_SLUG``` being ```autocrypt/autocrypt``` and ```TRAVIS_BRANCH``` being ```master```. Since conf.py is Python, it should be easy. You could change the logo or use another theme (derive from the normal one and change only one thing?).

dkg:
I'd just as soon not use travis-specific things, and have it all done directly from sphinx, so that my local "make html" covers the same ground.

dkg:
ok, that didn't appear to work.  i don't think i understand travis well enough to make this happen appropriately.  I've manually reverted the change in question and left it commented out in doc/conf.py in the hopes that someone else can figure this out.

dkg:
I figured it out and it works now!  thanks to @fschulze for pointing me in the right direction.

PR #299, opened 2017-12-20
Display "THIS IS A DRAFT" on everything but the master branch

dkg:
This is a change to the website, and is intended to help differentiate
proposed branches from the live (master) branch of the website.

Closes: #295 

You can see this effect live currently at https://staging.autocrypt.org/distinct-staging/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ addons:
 
 install:
   - pip install sphinx
+  - pip install gitpython
 
 script:
   - cd doc
--- a/doc/_static/staging.css
+++ b/doc/_static/staging.css
@@ -1,0 +1,19 @@
+body:before {
+  content: "THIS IS A DRAFT! please refer to https://autocrypt.org/ for the current Autocrypt website";
+  display: block;
+  position: fixed;
+  top: 0px;
+  left: 0px;
+  padding: 1em;
+  color: black;
+  font-weight: bold;
+  font-size: 150%;
+  min-height: 2em;
+  z-index: 10;
+  background-color: rgba(255,0,0,0.1);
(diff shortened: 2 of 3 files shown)


hpk42:
I think it's a bit too obnoxious. can you reduce its visual impact, make it smaller?  maybe make it appear at the very top, shifting everything down or so?  I understand it should b obnoxious and obvious but there are also good reasons to have branches and send links around so that people can look at new layout, new content .

fschulze:
I agree with hpk, it should be a banner on top to make it more readable but still obvious.

hpk42:
It would be nice if the "draft note" links to the appropriate github-branch page -- we could simply say: 
"You are viewing the 'branchname' draft, whereas official site is at: https://autocrypt.org"

and have the "branchname draft" thing link to github repo. 

not requiring it for more, just mentioning it if it's not a hassle.

dkg:
I've moved the banner to the top, instead of the old `position: fixed`.  I think this is dangerous, because it means someone looking at a device with a small screen who gets a link to (or scrolls to) an element deeper in the page will be unlikely to see it, but it's still better than nothing, and i'd like to get this merged so we can have some sort of visual distinction on https://staging.autocrypt.org/

I've also reduced the font size, as requested.

After ~10 minutes of fiddling, I couldn't see how to put the branchname in the warning, so i've left that for future work.

fschulze:
@dkg What about adding padding to the top of the body and make the :before absolutely positioned, so it's always at the top?

Besides that I like the look now.

dkg:
OK, i've moved it back to making it absolutely-positioned, and added padding at the top so that it doesn't always obscure the top of the page.  i think this is good to merge now.

dkg:
After some discussion on IRC, i'm going ahead with this merge.

PR #303, opened 2017-12-21
added c3 sessions to events

lefherz:
@hpk42 is it okay to have a call to action with an exclamation mark or is this army tone?


hpk42:
i might not use the "!" myself much don't object. thanks for caring for the sessions!

Issue #308, opened 2017-12-21
add press-team@autocrypt.org alias


dkg:
[i proposed a press team on the mailing list](https://lists.mayfirst.org/pipermail/autocrypt/2017-December/000289.html).

Could we set that up?  maybe we should also point `contact@` to the same list just in case someone is trying to write to it..

I know that #301 needs to be worked on --  i'm happy to work on it, and will transfer the alias when that happens.  but i hope this change doesn't need to block on #301.

dkg:
assigning to @hpk42 because i don't know who else has permissions to edit `/etc/aliases` on the current MX.

hpk42:
see my suggestion #312

hpk42:
the press-team at autocrypt.org is configured now with the members mentioned in the press-team doc.

PR #309, opened 2017-12-21
propose initial press team policy

dkg:


dkg:
fwiw, merging this into master doesen't mean the policy is perfect, it just means it's good enough to work with for now.  i hope we'll see changes as we figure out what works.

dkg:
See also #308

r10s:
looks good to me, i think this can be merged and refined later.

hpk42:
I started from your branch but diverged enough to do an alternative #312 
that "press-team" alias is currently matching what is in the #312. Note it's press-team, not "press"

lefherz:
I like #312, because it is text which you actually want to put on a website ;) 
But this one has a justification too, of course, transparency for the community.

Actually, I'd put #312 somewhere visible in contact.rst, and put a link below, which links to these #309 specifications, if anyone is interested they can find them.

dkg:
I'm closing this because i think #312 does everytihng this one does, but has more consensus around it.

PR #312, opened 2017-12-22
Press-team doc

hpk42:
This is an alternative to #309, differences

- it is not a policy but directly a doc where journalists may land
- it does not have "SHOULDs" but describes practises 
- it talks rather about accountability and not technical inbox-mechanics
- listed dkg, valodim, me as press-team members directly

### DIFF
--- a/doc/press-team.rst
+++ b/doc/press-team.rst
@@ -4,17 +4,20 @@ Autocrypt press team
 If you have any press-related inquiries please send e-mail to
 the `press-team <mailto:press-team@autocrypt.org>`_, The press team
 strives to answer within two workdays on press-related inquiries,
-and often will get back faster.
+and often will get back faster (we have a knack for e-mail).
 
 This document describes our practises how we handle and account for
 press communications.
 
-Initial team members
+Current team members
 --------------------
+
+- dkg
 
 - Holger Krekel / hpk42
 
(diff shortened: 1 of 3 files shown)


Valodim:
looks good to me. "two work days" sounds more reasonable than "24 hours"

r10s:
also looks good to me. 

did really laughing when reading "we have a knack for e-mail" :) **very good**

doc/press-team.rst
```diff
@@ -23,16 +23,19 @@ Current team members
...
-often journalists do not ask back before they publish articles, or do
-not even contact a project before reporting.
+a link of press articles to the community mailing list.  Anyone
+can ask how media reporting came about and how and who, if anyone, was
+involved from the press-team.  Note that often journalists do not ask
+back before they publish articles, or do not even contact a project
+before reporting.
+
+Press-team members remind each other when press-team internal
+discussions leak into community matters that are better discussed
```
Valodim: isn't it the opposite, i.e. "when community matters leak into press-team internal"?

hpk42: a bit ambigious but the remainder of the sentence clarifies it i think "that are better discussed through issues or mailing lists".  i  replaced "leak into" with "touch"  FWIW.

Issue #320, opened 2018-01-03
remove autocrypt impl repositories from organization 


hpk42:
while discussing https://github.com/autocrypt/py-autocrypt/pull/19 on IRC @azul brought up that maybe py-autocrypt could be better moved out of the org.  I tend to agree and suggest that we strive to generally have implementations repos move out of autocrypt and rather have autocrypt.org point to implementations which then live elsewhere.  Small scripts to generate test data and probably also ui-usability tests are fine to remain though.  

autocrypt is really about the spec and the group overall is not really prepared to collectively maintain software.

dkg:
It's fine by me if we want to move the implementations out of the organization.  certainly there's no one implementation that has all autocrypt contributors active in it.

dkg:
what else needs to be done to complete this?  should we delete the older "inbome" or other repositories that are basically legacy now?

azul:
To me deleting legacy repos or at least archiving them would make a lot of sense.

PR #321, opened 2018-01-04
address #320 -- shift py-autocrypt implementation to non-organization

hpk42:
address #320 -- prepare shift py-autocrypt implementation to non-organization location and mention juga0/pyac as another python supporting autocrypt effort. Accepting this merge be accompanied with moving the py-autocrypt repository.

### DIFF
--- a/doc/dev-status.rst
+++ b/doc/dev-status.rst
@@ -75,8 +75,6 @@ For developers
 
 Source code:
 
-- `py-autocrypt code <https://github.com/autocrypt/py-autocrypt>`_
-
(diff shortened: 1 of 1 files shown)


juga0:
I'm ok with moving out of org code repos.

This PR has a typo, missing the last ``_`` for the link. I'd also add link for pgpy, since is not well known.

The pyac line would be then:

    - `pyac <https://github.com/juga0/pyac>`_ (uses `PGPy <https://pgpy.readthedocs.io>`_)

juga0:
And i'd also add ``pyac`` to the table, as in #322

PR #322, opened 2018-01-04
Add pyac to dev status table

juga0:


doc/dev-status.rst
```diff
@@ -22,6 +22,8 @@ Last updated: ``2017-12-14``
 +--------------------------------------+--------+--------+--------+--------+--------------+--------+--------+--------+----------+
 |`py-autocrypt`_                       |✔       |✔       |✔       |✔       |✘             |✘       |✘       |✘       |✔         |
 +--------------------------------------+--------+--------+--------+--------+--------------+--------+--------+--------+----------+
+|`pyac`_                               |✔       |✔       |✔       |✔       |✔             |✔       |✔       |✔       |✔         |
```
hpk42: wow, surprised to see all checkmarks already :)  curious: How does pyac generate recommendations?  couldn't find something dealing with it in the source (but may have missed it) Can pyac already process ASMs from delta or enigmail?  And do you get gossip keys from the encrypted parts of emails?

juga0: > wow, surprised to see all checkmarks already :) curious: How does pyac generate recommendations? 

sorry, but copy/paste, fixed

> Can pyac already process ASMs from delta or enigmail?

i haven't checked, feel free to check it.
Maybe we should also add those to the tests in autocrypt repo?

> And do you get gossip keys from the encrypted parts of emails?

Yes: juga0/pyac/blob/master/autocrypt/message.py#L332

doc/dev-status.rst
```diff
@@ -22,6 +22,8 @@ Last updated: ``2017-12-14``
 +--------------------------------------+--------+--------+--------+--------+--------------+--------+--------+--------+----------+
 |`py-autocrypt`_                       |✔       |✔       |✔       |✔       |✘             |✘       |✘       |✘       |✔         |
 +--------------------------------------+--------+--------+--------+--------+--------------+--------+--------+--------+----------+
+|`pyac`_                               |✔       |✔       |✔       |✔       |✘             |✔       |✔       |✔       |✔         |
```
hpk42: IMHO the "setup message" tick should indicate things work in both directions with ASM messages (e.g. Delta.chat only says "started" even though delta can send ASMs to enigmail successfully already).  You could add a footnote like K-9 does for keygen or enigmail for header-inject that says it's untested wrt to ASM interactions with enigmail/k9/delta.

dkg: This sounds like a reasonable amendment to me; but if the columns are unclear, the legend should also be improved to make it less ambiguous.

hpk42: I agree with incrementally changing things, the table and legend needs more qualitifcation.

PR #331, opened 2018-01-07
incremental improvements on @valodim's hotfix last night, aiming for:

hpk42:
- brevity
- trying to be legible and useful for both end/expert users and
  security-minded hackers

we've had two public waves of interest (first heise, hackernews now) and i
guess we will keep changing/adapting things as tides role in,
and in particular when more mail apps with l1 support are released ...
I think "single-click opt-in encryption" is a useful catchphrase

Moreover, mentioning encrypted group communication and multi-device
support needs to be mentioned, as these are pain
points known to both existing users and developers.

I btw welcome the evolving practise of relatively quick merges --
if we evolve a practise that subsequent PRs can change things
equally easily :)

### DIFF
--- a/doc/_templates/customindex.html
+++ b/doc/_templates/customindex.html
@@ -40,22 +40,23 @@
   <div>
     <h2>What is Autocrypt?</h2>
     <p>
-      Autocrypt is a set of guidelines for convenient
-      end-to-end-encryption of e-mails. Developers are working on
-      bringing it to several different e-mail programs.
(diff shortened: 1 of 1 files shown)


r10s:
Looks good to me - things get clearer on each iteration :)  

@Valodim thank you for your action last night :)

Valodim:
This lost the bit about piggybacking on plaintext email. Given that almost every other effort to "bring email encryption to the masses" relies on yet more infrastructure to do this, that seemed important to me.

PR #335, opened 2018-01-08
be more specific and replace 'piggybacking'

karissa:
just in case there weren't enough changes in the past week, here's another two cents.

piggybacking is colloquial and non-native speakers with a lower level will likely have trouble with it.

### DIFF
--- a/doc/_templates/customindex.html
+++ b/doc/_templates/customindex.html
@@ -43,7 +43,7 @@
       Autocrypt is a set of guidelines for developers to achieve
       convenient end-to-end-encryption of e-mails.
       It specifies how e-mail programs negotiate encryption
-      capabilities keys using regular e-mails. 
+      capabilities and transfer keys using regular e-mails. 
     </p>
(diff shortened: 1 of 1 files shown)


Valodim:
I think hpk's use of "encryption capabilities" was a way to avoid using "keys" here :) your version also slightly loses the notion that the negotiation happens along regular emails, they are not specifically sent for this purpose.

How about this?:

> It specifies how e-mail programs can transparently negotiate encryption capabilities in regular e-mail communication.

karissa:
are keys not transferred? Happy to change it but it sounds like you're avoiding something that people will find out anyway

Valodim:
Keys are transferred, of course. But for users this is an implementation detail, and for techies it's an obvious implication from "negotiate encryption". Users don't need to know what a "key" is to be able to use Autocrypt, or at least that is one of our goals. :)

karissa:
ok removed the keys part.

PR #340, opened 2018-01-13
update favicon to a square, transparent with 3 resolutions

r10s:
The current favicon has a size of 100x84 which is an unusual size AFAIK. Updated the favicon to a square with 3 different resolutions: 16x16, 32x32 and 64x64; preview: https://share.riseup.net/#pOcGlUIOJYyI3DhWS6eUvw 

The background is transparent; unless the background is blue, this should be fine.

The favicon was generated from autocrypt_logo_alone_orig.png with http://realfavicongenerator.net


gfa:
LGMT

PR #341, opened 2018-01-13
Update Implementation status table

lefherz:
renamed py-autocrypt to muacrypt, added gossip column.

If some projects implemented more features in the meantime, this would be a good opportunity to add this here. Is there some progress I'm not aware of?


hpk42:
LGTM

PR #343, opened 2018-01-17
Updates information for K-9

reitzig:
K-9 Mail supports Autocrypt since release [5.400](https://github.com/k9mail/k-9/releases/tag/5.400).

Also, fixes a repository-internal deadlink.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -1,7 +1,7 @@
 Mail apps which support Autocrypt
 ================================================
 
-**Last update: Dec 6th 2017**
+**Last update: Jan 17th 2018**
 
 The following Autocrypt-enabled mail apps are available
 and/or under development:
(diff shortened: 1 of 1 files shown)


Valodim:
Autocrypt support is fairly experimental in K-9: The key choice algorithm is not up to date with the 1.0 spec, and it's missing gossip and autocrypt setup messages. Autocrypt is *not* just "putting keys in headers", so I'm a little uncomfortable announcing "Autocrypt support" without qualification like that.

reitzig:
That's too bad. I took the release notes on face value. My bad for not checking more closely.

Valodim:
Sorry about that :) It's hard to find a good granularity of information to give to users.

An update to actual 1.0 logic and gossip is in the [pipe](https://github.com/open-keychain/open-keychain/pull/2253) [line](https://github.com/k9mail/k-9/pull/3105) though. Autocrypt Setup Message will follow.

doc/install.rst
```diff
@@ -26,16 +26,16 @@ You can test it via `Nightly Build`_ today.
 
 .. _Enigmail: https://enigmail.net
 .. _Thunderbird: https://www.mozilla.org/en-US/thunderbird/
-.. _`Nightly Build`: install-autocrypt-enigmail.html
+.. _`Nightly Build`: install-autocrypt-enigmail.rst
```
r10s: Offtopic, however, as we want the link to work on the rendered homepage, the existing `.html` link is correct  (the rendered homepage has HTML-files, not reStructuredText-files). I do not know a solution that works on both, the rendered homepage and on the preview on github.

PR #344, opened 2018-01-17
Fix/pyac gossip and link

juga0:
Sorry i didn't see [#341](https://github.com/autocrypt/autocrypt/pull/341).
This fix `pyac` link and add it to the `gossip` column.


r10s:
Looks good to me (congratulations for the first gossip checkmark :)

Issue #346, opened 2018-01-22
Website: Change Setup Description for Enigmail


pbrunschwig:
I completed the GUI for enabling Autocrypt:
1. Autocrypt is now enabled by default
2. enabling/disabling Autocrypt is now a per-account setting, available in the Account Manager

Therefore the last bullet on https://autocrypt.org/install-autocrypt-enigmail.html is now obsolete. If at all, you could write that enabling/diabling Autocrypt and the "prefer-encrypt" setting is now available via:
Windows, macOS: Tools > Account Manager > OpenPGP Security
Linux: Edit > Account Manager > OpenPGP Security

azul:
@pbrunschwig yay! Thanks for turning it on by default :fireworks: :dancing_men:

r10s:
great that Autocrypt is enabled by default :)
/me also celebrating \o/ /o/ \o/

lefherz:
Merged :)

PR #347, opened 2018-01-22
Enigmail doesn't require interfering with the config anymore #346

lefherz:
addresses #346 

@pbrunschwig fine?

### DIFF
--- a/doc/install-autocrypt-enigmail.rst
+++ b/doc/install-autocrypt-enigmail.rst
@@ -16,7 +16,6 @@ You can test Autocrypt in Enigmail already today:
 
 - The following setup may require a restart of Thunderbird
 
-- Then, navigate to "Edit / Preferences / Advanced / Config Editor", read the warning and 
-  scroll down to "extensions.enigmail.autocryptMode": Double-click the entry and enter a `1`
+- Autocrypt is now running. You can configure it at: Account Manager > OpenPGP Security.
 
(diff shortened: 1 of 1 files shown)


doc/install-autocrypt-enigmail.rst
```diff
@@ -16,7 +16,6 @@ You can test Autocrypt in Enigmail already today:
 
 - The following setup may require a restart of Thunderbird
 
-- Then, navigate to "Edit / Preferences / Advanced / Config Editor", read the warning and 
-  scroll down to "extensions.enigmail.autocryptMode": Double-click the entry and enter a `1`
+- Autocrypt is now running. You can configure it at: Account Manager > OpenPGP Security.
```
hpk42: "is now running" sounds odd.  Shouldn't it just say "You can enable autocrypt here: Account Manager > OpenPGP Security" and maybe even mention the precise option? (Haven't installed the latest nightly yet so can't say what it is)

azul: It's more disable, right?
How about "Autocrypt is part of the default configuration. You can configure and disable it at:..." (If one can also configure it...)
Will install latest nightly tomorrow and try.

pbrunschwig: I agree with @azul, that sounds better.

lefherz: I'm fine with "Autocrypt is part of the default configuration", but we should remove the bullet point - it isn't part of the step-by-step guide anymore.

PR #349, opened 2018-01-23
fix examples entry point (a bit)

hpk42:
- rather than pointing to a 404 page (for a long time) just point to the level1 spec for the header examples
- strike the redundant "NOTE" which we already have similarly in the Level 1 spec intro page.
- remove "type=1" from headers in examples


r10s:
looks good to me; I also think, the note regarding active/passive attacks in the spec is clear and sufficient.

PR #371, opened 2018-03-05
a few paras at the beginning of the post, kind of

hpk42:
telling a different story/narrative than what currently is there.  Some of the points in my added paras are not made in the existing article.  I also think it's better to start off relatively concrete and go more abstract later if needed.  Maybe the rest of the article could be restructured a bit to incorporate this new beginning?  i can try to look again at things later today/tonight.


lefherz:
Good to have those changes in draft.rst, and I'll look what I can put into index.rst, too.

PR #374, opened 2018-03-13
initial draft on remailers

Valodim:
To have a basis for discussion, I created a braindump of my current thoughts on mailing lists. Most of these resulted from a number of discussions with @dkg, @hpk42  and @ge-fa.

Please comment :)

### DIFF
--- a/doc/remailers.rst
+++ b/doc/remailers.rst
@@ -14,15 +14,15 @@ For users with a typical intuition of end-to-end encryption, the expected end
 points of end-to-end encrypted mailing lists will likely be the individual
 subscribers. This is very difficult to achieve on top of existing e-mail
 infrastructure, since it requires either negotiation between all endpoints, or
-use of highly experimental encryption techniques such as homomorphic or
-attribute-based encryption.
+use of experimental encryption techniques such as homomorphic or attribute-based
+encryption.
 
 A compromise solution is to encrypt from the sender only to the remailer, which
 then decrypts the message and individually encrypts to all subscribers. This
-method strictly violates end-to-end confidentiality, since the remailer can
-access to the forwarded plaintext. However, the remailer will only need to
-handle sensitive data transiently, which means messages are secure as long as
-the remailer is honest at the time of sending.
+method violates strict end-to-end confidentiality, since the remailer can access
+to the forwarded plaintext. However, the remailer will only need to handle
+sensitive data transiently, which means messages are secure as long as the
+remailer is honest at the time of sending.
 
 This document focuses on seamless support for re-encryption based end-to-end
 security in existing remailers, in keeping with the more general Autocrypt
@@ -66,7 +66,8 @@ Injecting Reamiler Autocrypt-Headers
 
 To inform subscribers of its key, the remailer injects an Autocrypt-header into
 outgoing messages. The header format is the same as usual, where the ``addr``
-attribute is the address of the list, and ``keydata`` is their Autocrypt key.
+attribute is the address of the list, and ``keydata`` is the list's Autocrypt
+key.
 
 In addition to those two attributes, the remailer SHOULD add the
 ``_list-recommendation`` attribute, and MAY add ``_list-recipients`` attribute.
@@ -76,10 +77,13 @@ used by regular mailbox addresses. Both of these attributes are non-critical
 does not handle these attributes to skip over them, falling back to the usual
 Autocrypt behavior.
 
-For example::
+Example of Autocrypt header for remailer with opt-in encryption::
+
+    Autocrypt: addr=post@list.example; _list-recommendation=available; _list-recipients=7/15; keydata=XXX
+
+Example of Autocrypt header for remailer with mandatory encrypytion::
 
     Autocrypt: addr=post@list.example; _list-recommendation=mandatory; keydata=XXX
-    Autocrypt: addr=post@list.example; _list-recommendation=available; _list-recipients=7/15; keydata=XXX
 
 The ``_list-recommendation`` attribute SHOULD be included in an Autocrypt header
 sent by a remailer. Its value may be either ``available`` or ``mandatory``. If
--- a/doc/remailers.rst
+++ b/doc/remailers.rst
@@ -1,0 +1,125 @@
+Autocrypt and Mailing Lists
+====
+
+General Considerations
+----
+
+"Mailing lists" are a special use case of e-mail, where some address is managed
+not as an inbox that stores e-mail for retrieval, but instead forwards all
+messages to a configurable list of subscribers. For the case of end-to-end
+encryption, the introduction of a relaying agent brings up the question of what
+constitutes an "Endpoint".
+
+For users with a typical intuition of end-to-end encryption, the expected end
+points of end-to-end encrypted mailing lists will likely be the individual
+subscribers. This is very difficult to achieve on top of existing e-mail
+infrastructure, since it requires either negotiation between all endpoints, or
+use of highly experimental encryption techniques such as homomorphic or
+attribute-based encryption.
+
+A compromise solution is to encrypt from the sender only to the remailer, which
+then decrypts the message and individually encrypts to all subscribers. This
+method strictly violates end-to-end confidentiality, since the remailer can
+access to the forwarded plaintext. However, the remailer will only need to
+handle sensitive data transiently, which means messages are secure as long as
+the remailer is honest at the time of sending.
+
+This document focuses on seamless support for re-encryption based end-to-end
+security in existing remailers, in keeping with the more general Autocrypt
+approach.
+
+Autocrypt-enabled Remailer
+----
+
+An Autocrypt-enabled remailer maintains an Autocrypt key of its own. During
+regular operation, it collects Autocrypt-headers of all list participants via
+incoming mail, to keep track of their keys. Then, depending on the availability
+of encryption keys for participants, the remailer injects its own
+Autocrypt-headers into outgoing mail to signal availability of encryption.
+
+More specifically, an Autocrypt-enabled mailing list adds three behaviors to its
+remailing logic:
+
+1) For all incoming messages, keep track of Autocrypt-headers. This mechanism is
+   the same that is used by clients, and requires no modifications.
+2) Inject an Autocrypt-header into outgoing messages. This header serves the
+   same purpose as usual, but adds some details for the mailing list use case.
+   The header might only be injected under certain conditions, e.g. if
+   encryption keys are available for a certain number of subscribers.
+3) When receiving an encrypted message, decrypt it and re-encrypt for individual
+   delivery to each Autocrypt-enabled subscriber. Subscribers for whom no keys
+   are available do not receive the message, but optionally a notification of
+   this circumstance (see below).
+
+The presented mechanism works well for relatively small groups of mostly active
+and Autocrypt-enabled subscribers, but becomes increasingly ineffective with
+larger numbers of passive or non Autocrypt-enabled subscribers. As with any
+encrypted group communication, the benefits of end-to-end encryption also dilute
+with larger numbers of recipients.
+
+For compatibility with clients that are not Autocrypt-capable, the remailer
+software might want to offer manual configuration of keys, e.g. via a web
+interface.
+
+Injecting Reamiler Autocrypt-Headers
+~~~~~
+
+To inform subscribers of its key, the remailer injects an Autocrypt-header into
+outgoing messages. The header format is the same as usual, where the ``addr``
+attribute is the address of the list, and ``keydata`` is their Autocrypt key.
+
+In addition to those two attributes, the remailer SHOULD add the
+``_list-recommendation`` attribute, and MAY add ``_list-recipients`` attribute.
+Both of these attributes are exclusive for use by remailers, and MUST NOT be
+used by regular mailbox addresses. Both of these attributes are non-critical
+(starting with an underscore), which will cause an Autocrypt-capable client that
+does not handle these attributes to skip over them, falling back to the usual
+Autocrypt behavior.
+
+For example::
+
+    Autocrypt: addr=post@list.example; _list-recommendation=mandatory; keydata=XXX
+    Autocrypt: addr=post@list.example; _list-recommendation=available; _list-recipients=7/15; keydata=XXX
+
+The ``_list-recommendation`` attribute SHOULD be included in an Autocrypt header
+sent by a remailer. Its value may be either ``available`` or ``mandatory``. If
+the value is set to ``mandatory``, the remailer requires encryption of every
+message, and will reject plaintext messages. If the value is ``available``,
+encryption is offered as an opt-in mechanism and plaintext messages will be
+handled as usual. In either case, encrypted messages will be delivered
+exclusively to subscribers for whom an encryption key is available.
+
+To inform the decision of whether to encrypt or not, the remailer MAY include a
+``_list-recipients`` attribute. This attribute contains the current number of
+subscribers for which an encryption key is available, and the total number of
+subscribers, separated by a slash character ('/'). This attribute is only useful
+if the ``_list-recommendation`` attribute is set to ``available``.
+
+Administrative Considerations
+----
+
+Graceful Up- and Downgrade
+~~~
+
+To allow graceful up- and downgrade of lists with no intervention by the list
+administrator, a remailer software distribution could automatically include
+Autocrypt-headers with outgoing mail depending on the number of
+Autocrypt-enabled subscribers. For example, a remailer software distribution may
+choose to inject Autocrypt headers only if encryption keys are available for a
+majority of subscribers. Note that obviously, end-to-end encryption is
+unsuitable for lists that are publicly archived.
+
+Missed Message Notifications
+~~~
+
+A remailer MAY send notifications to recipients that were unable to receive an
+encrypted message, if their key was unavailable or unusable for some reason. If
+available, a digest mechanism can be used to aggregate this information and
+avoid redundant notifications.
+
+Limiting Plaintext Exposure
+~~~
+
+An honest remailer SHOULD limit exposure to the plaintext data by re-using the
+session key of the original message and re-encrypting it directly, to avoid
+having the actual plaintext in memory at any time.


ge-fa:
Thanks for pushing this!

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+Autocrypt behavior.
+
+For example::
+
+    Autocrypt: addr=post@list.example; _list-recommendation=mandatory; keydata=XXX
+    Autocrypt: addr=post@list.example; _list-recommendation=available; _list-recipients=7/15; keydata=XXX
+
+The ``_list-recommendation`` attribute SHOULD be included in an Autocrypt header
+sent by a remailer. Its value may be either ``available`` or ``mandatory``. If
+the value is set to ``mandatory``, the remailer requires encryption of every
```
Valodim: Note that this introduces a `mandatory` mode to Autocrypt. I consider this a notably different use case than a unilateral preference for encryption by default, because it implies that messages that aren't encrypted really strictly are not going to be delivered. There should probably be an override, but it's a nice usability feature to have an immediate warning for fully encrypted lists.

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+encryption is offered as an opt-in mechanism and plaintext messages will be
+handled as usual. In either case, encrypted messages will be delivered
+exclusively to subscribers for whom an encryption key is available.
+
+To inform the decision of whether to encrypt or not, the remailer MAY include a
+``_list-recipients`` attribute. This attribute contains the current number of
+subscribers for which an encryption key is available, and the total number of
+subscribers, separated by a slash character ('/'). This attribute is only useful
+if the ``_list-recommendation`` attribute is set to ``available``.
+
```
Valodim: So far there are no technical details on how client-side header processing is affected by this. I wonder how much sense it makes to express this in terms of the main spec's `peer_state[]` info, introducing new autocrypt-recommendations, etc?

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+The ``_list-recommendation`` attribute SHOULD be included in an Autocrypt header
+sent by a remailer. Its value may be either ``available`` or ``mandatory``. If
+the value is set to ``mandatory``, the remailer requires encryption of every
+message, and will reject plaintext messages. If the value is ``available``,
+encryption is offered as an opt-in mechanism and plaintext messages will be
+handled as usual. In either case, encrypted messages will be delivered
+exclusively to subscribers for whom an encryption key is available.
+
+To inform the decision of whether to encrypt or not, the remailer MAY include a
+``_list-recipients`` attribute. This attribute contains the current number of
```
Valodim: There is a tradeoff in granularity of specification here. Would it be helpful to have consistency in similar terms to the `discourage` state? I feel like just leaving it up to the implementation what to make of this info is not giving enough guidance

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+~~~~~
+
+To inform subscribers of its key, the remailer injects an Autocrypt-header into
+outgoing messages. The header format is the same as usual, where the ``addr``
+attribute is the address of the list, and ``keydata`` is their Autocrypt key.
+
+In addition to those two attributes, the remailer SHOULD add the
+``_list-recommendation`` attribute, and MAY add ``_list-recipients`` attribute.
+Both of these attributes are exclusive for use by remailers, and MUST NOT be
+used by regular mailbox addresses. Both of these attributes are non-critical
```
Valodim: should this be critical? I feel like they shouldn't, but there may be arguments for a yes here.

Valodim:
made some inline comments for points that others might want to address.

one more thought: would it be helpful to have a "key request" mechanism? e.g. an indication in the header that some user requests a key?

Roughly brainstorming:

> `Autocrypt-Request: true`

A message that includes this header could be interpreted by a MUA and offer single-click replying with an empty mail (that includes an Autocrypt header), and content that asks the user to do so manually as a fallback. This would give a clear workflow to ask someone for their keys if the user subscribes to a mailing list (or other service), similar to an "invite to chat" mechanism from messengers. This might even be useful in a more general context, but mailing lists seem like a good candidate to gather user feedback on how well this works.

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+not as an inbox that stores e-mail for retrieval, but instead forwards all
+messages to a configurable list of subscribers. For the case of end-to-end
+encryption, the introduction of a relaying agent brings up the question of what
+constitutes an "Endpoint".
+
+For users with a typical intuition of end-to-end encryption, the expected end
+points of end-to-end encrypted mailing lists will likely be the individual
+subscribers. This is very difficult to achieve on top of existing e-mail
+infrastructure, since it requires either negotiation between all endpoints, or
+use of highly experimental encryption techniques such as homomorphic or
```
dkg: drop "highly" -- it's bad enough that they're experimental :)

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+"Mailing lists" are a special use case of e-mail, where some address is managed
+not as an inbox that stores e-mail for retrieval, but instead forwards all
+messages to a configurable list of subscribers. For the case of end-to-end
+encryption, the introduction of a relaying agent brings up the question of what
+constitutes an "Endpoint".
+
+For users with a typical intuition of end-to-end encryption, the expected end
+points of end-to-end encrypted mailing lists will likely be the individual
+subscribers. This is very difficult to achieve on top of existing e-mail
+infrastructure, since it requires either negotiation between all endpoints, or
```
dkg: "negotiation between all endpoints" should say something like: e.g., an explicitly Cc'ed list of recipients, with everyone doing "reply all".

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+use of highly experimental encryption techniques such as homomorphic or
+attribute-based encryption.
+
+A compromise solution is to encrypt from the sender only to the remailer, which
+then decrypts the message and individually encrypts to all subscribers. This
+method strictly violates end-to-end confidentiality, since the remailer can
```
dkg: should be "this method violates strict end-to-end confidentiality"

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+access to the forwarded plaintext. However, the remailer will only need to
+handle sensitive data transiently, which means messages are secure as long as
+the remailer is honest at the time of sending.
+
```
dkg: This analysis doesn't take into account future key compromise.  And adversary who captures a message while the remailer is honest ("at time of sending") and then later steals the remailer's decryption key can violate the confidentiality of the message.

This can be ameliorated with subkey rotation and other schemes, but if we don't go into that here, we're handwaving away a legit concern.

Another way that i've seen this framed is to say that a an encrypted remailer-style mailing list with n subscribers has n+1 points of vulnerability for each message.

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+software might want to offer manual configuration of keys, e.g. via a web
+interface.
+
+Injecting Reamiler Autocrypt-Headers
+~~~~~
+
+To inform subscribers of its key, the remailer injects an Autocrypt-header into
+outgoing messages. The header format is the same as usual, where the ``addr``
+attribute is the address of the list, and ``keydata`` is their Autocrypt key.
+
```
dkg: should be "and ``keydata`` is the list's Autocrypt key"

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+used by regular mailbox addresses. Both of these attributes are non-critical
+(starting with an underscore), which will cause an Autocrypt-capable client that
+does not handle these attributes to skip over them, falling back to the usual
+Autocrypt behavior.
+
+For example::
+
+    Autocrypt: addr=post@list.example; _list-recommendation=mandatory; keydata=XXX
+    Autocrypt: addr=post@list.example; _list-recommendation=available; _list-recipients=7/15; keydata=XXX
+
```
dkg: These examples should be separated.  without separation, it might imply that the same header is being set twice in a single message.

doc/remailers.rst
dkg: do we want to include this here?  It's conceivable that this "mandatory" mode could be relevant for non-list mailboxes as well, where cleartext messages are refused delivery.  if so, it'd be strange to have two ways to express it.

I recommend splitting the idea of ``mandatory`` out into a separate/orthogonal discussion, and removing `_list-recommndation`` from this draft.

doc/remailers.rst
dkg: yes, i agree that we should expand `peer_state[listaddr]` explicitly if we're going to document this.

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+~~~
+
+To allow graceful up- and downgrade of lists with no intervention by the list
+administrator, a remailer software distribution could automatically include
+Autocrypt-headers with outgoing mail depending on the number of
+Autocrypt-enabled subscribers. For example, a remailer software distribution may
+choose to inject Autocrypt headers only if encryption keys are available for a
+majority of subscribers. Note that obviously, end-to-end encryption is
+unsuitable for lists that are publicly archived.
+
```
dkg: Maybe add an additional caveat here about wanting some hysteresis or other kinds of smoothing -- one person's key becoming unavailable and then available shouldn't cause flip-flopping of the presence of the header on subsequent messages.

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+Missed Message Notifications
+~~~
+
+A remailer MAY send notifications to recipients that were unable to receive an
+encrypted message, if their key was unavailable or unusable for some reason. If
+available, a digest mechanism can be used to aggregate this information and
+avoid redundant notifications.
+
```
dkg: I like this suggestion about notifications, but i don't think it goes far enough.

It should also discuss the remailer sending notifications to the message *sender* that some number of recipients were unable to receive a given cleartext message?

doc/remailers.rst
```diff
@@ -0,0 +1,125 @@
...
+Limiting Plaintext Exposure
+~~~
+
+An honest remailer SHOULD limit exposure to the plaintext data by re-using the
+session key of the original message and re-encrypting it directly, to avoid
+having the actual plaintext in memory at any time.
```
dkg: This implies a few things that i think are interesting:

 * the message signature will come from the initial sender, *not* from the mailing list
 * the re-mailer cannot tamper with the content of the message (e.g. no adding footers, no "malware" checking)
 * any internally-protected Autocrypt headers (e.g. "memory-hole") might override whatever Autocrypt headers the re-mailer itself injects.

I'm not sure about the logistical consequences of these ideas.

dkg:
thanks for this work, vincent!

I'm not convinced that we want to experiment with general features like
"request to chat"/Autocrypt-Request on the mailing list itself -- these
seem like things we should approach orthogonally, and not try to sneak
them in during mailing list proposals.

Many mailing lists have an e-mail driven subscribe/confirm workflow --
and the confirmation request itself often comes from the mailing list.
Those messages represent an opportunity to seed mailing list keys,
without introducing a new autocrypt-request standard.

        --dkg

doc/remailers.rst
Valodim: This can get annoying fast. I also think that if the client is autocrypt-mailing-list-capable, the `_list-recipients` attribute is a much nicer way to achieve this a priori. A notification to the sneder would be more of a fallback for non-autocrypt clients I guess?

doc/remailers.rst
Valodim: It doesn't actually flip-flop, unless there is a 35 days delay between the messages (and even then it only becomes "discouraged") :)

doc/remailers.rst
Valodim: I don't think it would be strange at all, because there is value in displaying a specific message to the user about the actual case. Even if the technical implication is the same or similar, it feels like a good idea to me to keep the use cases apart on the protocol layer, otherwise implementations *will* conflate them. "This recipient is a fully encrypted mailing list, unencrypted messages may not be delivered."

This is also the reason for the `MUST NOT` wrt using those attributes for non-mailinglist addresses.

doc/remailers.rst
Valodim: Heh, I was kind of aware but didn't know how to put it nicely without being too negative about the whole thing. The n+1 thing does sound like a good framing

doc/remailers.rst
Valodim: Hm. My thought was that mailing lists can still be used for delivery, it's just the e2e crypto that requires negotiation between all endpoints. A Cc'ed list of recipients should be distinctly *not* difficult to achieve, given `Autocrypt-Gossip` :) should make that point clearer

doc/remailers.rst
dkg: well, _list-recipients is what's true at *last message recieved* time, but during actual new message delivery, the state might be different.

"this can get annoying fast" is probably true, but it's also true for the non-crypto recipients.  we can mitigate that problem with the same "digest mechanism" idea, right?

doc/remailers.rst
dkg: it could flip flop if a new user joins, or if a user quits and the number generated is based on a proportion of subscribers, right?  or a new user subscribes (without an autocrypt key) and then later adds their autocrypt key.  

My point is that we ought to explicitly think about all of the use cases and ensure that we are ok with this kind of dynamic threshhold-setting.

alternately, we could just emit the list's autocrypt header as long as the remailer thinks at least one recipient is autocrypt-capable and let the clients decide based on the `_list-recommendation`.

if we really want to ensure that peers only send to mailing lists if they can deal with this kind of decision-making, we could also convert `_list-recommendation` to `list-recommendation` (making it critical, so that normal autocrypt clients would ignore the header until they've been taught about encrypted re-mailers).

doc/remailers.rst
Valodim: I think last message receive time is fairly good info on mailing lists that aren't very inactive, and the rough impression will still be right. Having both mechanisms seems redundant, and if I had to choose the header seems preferable.

> "this can get annoying fast" is probably true, but it's also true for the non-crypto recipients. we can mitigate that problem with the same "digest mechanism" idea, right?

If we wanted to go this route, a limit of one message every X hours/days would make more sense, for immediate feedback when a message is sent, without being spammy for multiple messages (and the info is not going to change much between batches sent).

doc/remailers.rst
dkg: I'm still not convinced, sorry.  two functionally equivalent specs *will* be conflated by implementations even with a `MUST NOT` that tries to scare the implementers off.

and `_list-recommendation` isn't the only way to flag the list as a mailing list -- we can use `_list-recipients` (or a critical `list-recipients`, if we want to require the MUA to be able to know about this breach of e2e to present it as encryption-capable to the user).

Bundling the concept of "cleartext mail will not be delivered" with "you are sending to a mailing list" really doesn't seem like the right way to go.

doc/remailers.rst
dkg: see comments elsewhere in this discussion for an argument for why `list-recipients` might want to be critical.

doc/remailers.rst
Valodim: I meant that while inclusion of the header will flip-flop, the calculated autocrypt-recommendation on clients will not due to the way timestamps are handled. If it didn't that would be annoying for the user even for the non-mailinglist case. Is this included in your consideration?

doc/remailers.rst
Valodim: The `mandatory` bit was not included in my first draft, I added it based on feedback by @ge-fa for the specific use case of Schleuder. Perhaps he can also weigh in on this.

> Bundling the concept of "cleartext mail will not be delivered" with "you are sending to a mailing list" really doesn't seem like the right way to go.

There is an important difference here: For mailing lists, messages *actually won't be delivered*. This is not the case for regular mailboxes, "I will disregard plaintext messages you send to me" is not something that actually works with a human backend.

I do think the `mandatory` mode is useful, but I'm still sceptical if we should design this for a more general use case that we don't have a clear picture of and can't really reason about. Do you have another use case in mind?

A more general take leads to `prefer-encrypt=mandatory`, but honestly the abuse potential of such a setting is so high that I'd rather not have it in Autocrypt.

Valodim:
anybody else wanna weigh in here? @azul @ge-fa @hpk42?

@dkg and I could probably just go on here, but I'd prefer to have a better feeling of what the consensus is on some of these aspects.

doc/remailers.rst
azul: @dkg s first point probably also means that clients who are not aware of this may display the message as having a broken signature.

In general i think a signature by the initial sender with a intended-recipients header would provide the best assurance of integrity. It would signal who the message was send by and that it was indeed intended for the list. However the logic in evaluating such a message seems quite complex - in particular with the headers the remailer has to change for delivery.

azul:
I would be very happy to see this merged as a summary of the current state of discussion. Discussing refinements later will be easier than discussing everything together in a single PR.

ge-fa:
Sorry for the delay, I saw the question(s) quite some time ago, but wasn't able to react, due to too much work. :( I'll have a look tomorrow.

Valodim:
@ge-fa ping :)

doc/remailers.rst
```diff
@@ -0,0 +1,129 @@
...
+1) For all incoming messages, keep track of Autocrypt-headers. This mechanism is
+   the same that is used by clients, and requires no modifications.
+2) Inject an Autocrypt-header into outgoing messages. This header serves the
+   same purpose as usual, but adds some details for the mailing list use case.
+   The header might only be injected under certain conditions, e.g. if
+   encryption keys are available for a certain number of subscribers.
+3) When receiving an encrypted message, decrypt it and re-encrypt for individual
+   delivery to each Autocrypt-enabled subscriber. Subscribers for whom no keys
+   are available do not receive the message, but optionally a notification of
+   this circumstance (see below).
```
hpk42: This constitutes for me a particular policy of running a mailing list, certainly an interesting one.  It means that users who are autocrypt enabled can send each other confidential information, e.g. roomcodes for an event, passwords for sites , etc.

hpk42:
I appreciate your engagement and the current state of the document, very useful!  

Could you describe/scope things to be about a particular policy/mode of running a ML ("confidential sub-channel" mode or so)?  

One thing i'd like to achieve with muacrypt and mailman integration is that it also works with Delta.chat, i.e. delta users can subscribe/participate in a group whose subscriber list and message relay is managed via a remailer. As Delta does not offer an "encrypt" checkbox, there is no explicit choice about encryption, the setting is "default autocrypt opportunistic".  An "opportunistic mode"  of a remailer would send all messages to everyone, encrypting as determined by autocrypt default policies.  It does violate the "end-to-end" encryption people might expect but that is also true for the "encrypted sub-channel" mode. I understand that you make a distinction in that the "sub-channel" mode maintains at least end-to-end encryption between all end-points if we include the re-mailer as an end-point.  The "opportunistic" mode violates this -- a user encrypts but the remailer also sends it out in cleartext.  But if the user does not explicitely make a decision to encrypt, it just means we are "replacing cleartext with encrypted messages". 

FWIW If we want to maintain "end-to-end" expectations i see two possibilities: 

- proxy re-cryption 

- making the remailer and autocrypt end points exchange messages to allow each peer to encrypt to all the keys of members but not to any re-mailer key. 

Would be very happy if can you take up some of considerations and open up the doc a bit.  This all warrants a real-life discussion which i hope we are having soon (ASFAIK we are going to hear from Mozilla soon regarding the "Opportunistically encrypted mailing lists with Mailman and Autocrypt" which would help organizing and funding a topical gathering.

hpk42:
Two questions: 

1. what users expect to happen when they send a mail to a mailing list likely depends on if encryption was a choice.  If there was no choice and there is no UI indication of encryption displayed with messages, why not replace cleartext with encrypted messages wherever we can?  

2. do you see this PR as establishing a shared working document or do you already want to prescribe a policy?

hpk42:
Update: i am fine to merge this PR as i iff it gets a clear note at the beginning about its discussion-status  linking back to this PR #374. So it's recognized as a draft stating and referencing ongoing discussions.

dkg:
is it better to merge this, or to keep it as an unresolved branch?  given the number of open questions that the discussion has, i'm not convinced that merging it makes sense yet, even with a disclamatory header, but i might not understand the level of disclaimer you're talking about adding.

@hpk42 if you think a header might make this more palatable for a merge, maybe update the branch to include the proposed header?

doc/remailers.rst
```diff
@@ -0,0 +1,129 @@
+Autocrypt and Mailing Lists
+====
+
```
hpk42: NOTE: this document and the accompanying discussions on the https://github.com/autocrypt/autocrypt/pull/374  according pull request reflect the unresolved state of discussions in 2018 around the topic of remailers and Autocrypt integration.  Several people intend to revisit the interesting question of autocrypt/remailer integration when the times comes for it.  If you'd like to do a project towards this end please try engage through Autocrypt channels (IRC and ML at this point, as well as physical gatherings from time to time, also in 2019).

doc/remailers.rst
```diff
@@ -0,0 +1,129 @@
...
+General Considerations
+----
```
hpk42:
```suggestion
----NOTE: this document and the accompanying discussions on the #374 according pull request reflect the unresolved state of discussions in 2018 around the topic of remailers and Autocrypt integration. Several people intend to revisit the interesting question of autocrypt/remailer integration when the times comes for it. If you'd like to do a project towards this end please try engage through Autocrypt channels (IRC and ML at this point, as well as physical gatherings from time to time, also in 2019).
```

Valodim:
I was initially hopeful to get some more input here from mailing list folks after fruitful discussions at iff list year. But this isn't a use case I'm very actively thinking about right now, so I'm good with leaving this unresolved or as a wip marked document.

Maybe we can introduce a "drafts" area for experimental stuff we are working on, to encourage capturing thoughts in this style and allow for iterative pull request workflows?

Valodim:
Closing as stale. This is still a good start if support in mailing lists ever comes up again

PR #376, opened 2018-04-23
draft for discussion: Autocrypt-Draft-State header

Valodim:
Hey there,

following up on [this thread](https://lists.mayfirst.org/pipermail/autocrypt/2018-April/000325.html) on the mailing list, I wrote a proposal with my current thoughts on this, mostly to have something concrete to discuss and be able to comment line-wise.

A couple of points:

* I went with an Autocrypt specific header, rather than an OpenPGP one. The main reason for this is that a generic OpenPGP header would have to operate in a much broader and less understood context. It's also convenient to be able to reuse the autocrypt header attribute-list format with critical and non-critical attributes.

* The header is designed to be minimal and compatible: The most simple implementation just needs to parse whether "encrypt" is yes or no. Building on that, they can pick up the non-critical attributes.

* This is so far missing a mention of additional possible attributes `_sign-only` and `_encrypt-only`. We don't treat those states in Autocrypt itself, and clients can simply use their own non-critical attributes to store custom data. It still makes sense to mention those two (in a note, perhaps?) with a short description of their semantics, to avoid implementations coming up with different names for them.

* I'd like to have an attribute `_pgp-inline`. In K-9, when replying to a pgp/inline message, the reply will automatically be sent as pgp/inline as well, which is important for compatibility with some MUAs, particularly browser extensions. However, I noticed that Autocrypt actually says messages MUST be sent as PGP/MIME. We should probably reopen that discussion, possibly turn that into a SHOULD and give some guidance on how to handle PGP/INLINE rather than pushing clients into breaking an impractical MUST.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -771,10 +771,11 @@ all of which have a binary value of ``yes`` or ``no``:
 * The critical ``encrypt`` attribute specifies whether the message would
   be sent encrypted or not, at the time of saving the draft. It MUST
   be present in the header.
-* The non-critical ``_is-reply`` attribute indicates whether the message
-  is composed as a reply to an encrypted message. This affects the
-  :ref:`Autocrypt recommendation<final-recommendation-phase>`. It is
-  optional and defaults to ``no``.
+* The non-critical ``_is-reply-to-encrypted`` attribute indicates
+  whether the message is composed as a reply to an encrypted message.
+  This affects the :ref:`Autocrypt
+  recommendation<final-recommendation-phase>`. It is optional and
+  defaults to ``no``.
 * The non-critical ``_by-choice`` attribute indicates whether the value
   of the ``encrypt`` attribute was made by immediate user choice. It is
   optional and defaults to ``no``.
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -710,6 +710,8 @@ To avoid leaking metadata about a third party in the clear, an
 ``Autocrypt-Gossip`` header SHOULD NOT be added outside an encrypted
 MIME part.
 
+.. _key-gossip-update:
+
 Updating Autocrypt Peer State from Key Gossip
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -778,6 +780,26 @@ all of which have a binary value of `yes` or `no`:
 Example::
 
     Autocrypt-Draft-State: encrypt=yes; _by-choice=yes;
+
+Autocrypt Gossip headers for drafts
++++++++++++++++++++++++++++++++++++
+
+When opening a draft that indicates it should be encrypted when sent
+via the ``Autocrypt-Draft-State`` header, the resuming MUA might be
+lacking keys for recipients that were originally available when the
+draft was stored.
+
+To solve this problem, a MUA MAY make use of the `Autocrypt
+Gossip<autocryt-gossip>` mechanism to include the keys of intended
+recipients in the draft message. To do so, it simply places one
+``Autocrypt-Gossip`` header per recipient in the MIME header of the
+encrypted payload of the draft.
+
+Conversely, when opening a draft, a MUA MAY import keys from
+``Autocrypt-Gossip`` headers that are present in the MIME headers of
+the encrypted payload.  These headers should update the ``gossip_key``
+and ``gossip_timestamp`` fields of the relevant peers `as
+usual<autocryt-gossip-update>`.
 
 .. _account-management:
 
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -735,7 +735,7 @@ way:
    ``keydata`` attribute.
 
 Message Drafts
--------------
+--------------
 
 During message composition, the user may want to close the dialog and
 resume composition at a later point.  To preserve the session state,
@@ -783,6 +783,11 @@ all of which have a binary value of ``yes`` or ``no``:
 Example::
 
     Autocrypt-Draft-State: encrypt=yes; _by-choice=yes;
+
+The semantics of this header are defined only in the context of
+a message that is loaded as a draft. It SHOULD NOT be processed for
+messages loaded in any other context. It MUST be stripped from
+a message before it is sent.
 
 .. note::
 
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -758,25 +758,25 @@ Storing Draft State
 +++++++++++++++++++
 
 To store information about whether a message should be encrypted when
-sent, an `Autocrypt-Draft-State` header MAY be added to the draft when
+sent, an ``Autocrypt-Draft-State`` header MAY be added to the draft when
 it is stored.
 
-The `Autocrypt-Draft-State` header consists of a list of attributes
+The ``Autocrypt-Draft-State`` header consists of a list of attributes
 with the same syntax as the :ref:`Autocrypt header<autocrypt-header>`
 itself, which also supports critical and non-critical attributes
 following the same semantics.  There are three defined attributes,
-all of which have a binary value of `yes` or `no`:
+all of which have a binary value of ``yes`` or ``no``:
 
-* The critical `encrypt` attribute specifies whether the message would
+* The critical ``encrypt`` attribute specifies whether the message would
   be sent encrypted or not, at the time of saving the draft. It MUST
   be present in the header.
-* The non-critical `_is-reply` attribute indicates whether the message
+* The non-critical ``_is-reply`` attribute indicates whether the message
   is composed as a reply to an encrypted message. This affects the
   :ref:`Autocrypt recommendation<final-recommendation-phase>`. It is
-  optional and defaults to `no`.
-* The non-critical `_by-choice` attribute indicates whether the value
-  of the `encrypt` attribute was made by immediate user choice. It is
-  optional and defaults to `no`.
+  optional and defaults to ``no``.
+* The non-critical ``_by-choice`` attribute indicates whether the value
+  of the ``encrypt`` attribute was made by immediate user choice. It is
+  optional and defaults to ``no``.
 
 Example::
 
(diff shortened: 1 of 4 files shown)


Valodim:
Ah, almost forgot: We will also have to think a little about versioning of the spec: While this would be a non-breaking change, it certainly warrants a version 1.1, and when we do that we should have a changelog and archive of older versions. I noticed that people do pay attention to those versions, and the progress from one to another.

Valodim:
I noticed that we can use gossip headers to make sure all keys of intended recipients are available when a message is resumed up from a draft. I think that's a very useful feature to have, simple to implement it gossip is alrady available, and trivial to spec. I added a section about that.

pbrunschwig:
@Valodim I like your proposal, but there is one point I'd like to discuss. Why should we need to put the Autocrypt-Draft-State header in the encrypted payload? The message is only stored on your the IMAP server of your provider. There is not much value for the provider to know if the user set a specific encryption/signing state for the message. Thus, the main risk I see is that your provider could alter the header. But seriously, if you don't trust the provider on this level, then you better change your provider ...

Valodim:
Good question, I actually had to think why I did it like this for a bit :)

So it's two parts: For one, I thought it simply doesn't hurt. We know that the draft is not going to be useful for a client that doesn't know how to decrypt it, and at that point we might as well stuff as much metadata in the encrypted part as possible. Storing the subject encrypted in the payload is particularly useful, and I taught K-9 to do that shortly before writing this spec, so it seemed natural. Does this introduce difficulties for you?

The second reason is that headers in the encrypted payload are less likely to be included as-is when the message is prepared for delivery by clients that don't know what it is. This seemed like a general risk with proprietary headers in drafts that aren't meant for sending, but I don't actually know if any MUAs actually send headers they don't know (I think K-9 doesn't)

pbrunschwig:
This decision doesn't just introduce additional difficulties for me - it's simply not possible in Thunderbird. 

When editing a message (be it a reply or editing a draft), then Thunderbird does not provide access to _any_ of the original message data, except for the outmost headers. I don't see the original message structure at all, nor headers of any sub-parts at all. 

The only workaround I have is at message reading - this allows me to set some properties that I can later read when editing the message. But that's not guaranteed for editing drafts (they don't need to be read before being edited).

Valodim:
I don't understand. How can you do encrypted drafts then? Do you have access to the body?

Valodim:
@pbrunschwig ping. this is blocking my draft PR in k-9, some insight would be much appreciated :) I also couldn't get encrypted drafts to work interoperably with enigmail, perhaps this is related.

pbrunschwig:
@Valodim Sorry, I was on holidays and didn't notice you replied.

Saving a draft works exactly the same way as sending a message, with the small difference that the delivery type is "Save as draft" instead of "Send now" or "Send later". I capture the "save" or "send" event in the GUI, add replace one of the Thunderbird objects with my own object. This way, when I'm called later during send/save, I know what to do. Not very efficient... ;-)

Valodim:
So how do you restore state from a draft? Is there even a way I can save drafts to the imap sent folder that you can process?

pbrunschwig:
I have access to the top-level headers and to the message database. When you view a message with an encrypted subject, the subject in the message database is replaced with the original subject. But the database has only a limited set of fields, I can't add anything I'd like.

r10s:
In general, this looks good to me. 

Intuitively i would also have put the Autocrypt-Draft-State header in the encrypted payload, same as the subject and maybe other headers. But i understand that this is difficult in existing MUAs.

Valodim:
It's easy to do in K-9, and I actually just reverted the ten lines commit that did it :) but if it's not feasible in enigmail, the gain is small enough that it's an easy tradeoff.

r10s:
yes, i know. i wanted to write "But i understand that this might be difficult in some existing MUAs." :)

doc/level1.rst
```diff
@@ -710,6 +701,79 @@ way:
...
+provider, the MUA can encrypt the draft message to itself (but not the
+recipients) before uploading.  It MUST encrypt in this way if the
+message is going to be sent encrypted, and SHOULD encrypt if the user
+might decide to encrypt at a later point, or otherwise keep the draft
+local only.  Drafts encrypted in this way SHOULD be stored in PGP/MIME
+format.
+
+To allow storage of drafts independently from the user's secret key,
+encrypted drafts do not need to be signed.  Conversely, a MUA SHOULD
+ignore the signature status of encrypted drafts.
```
dkg: This makes it difficult to defend against draft injection by unauthorized clients.  I know that's an active attack and therefore out of scope, but it's not clear what we gain by ignoring this status.

doc/level1.rst
```diff
@@ -710,6 +701,79 @@ way:
...
+
+Example::
+
+    Autocrypt-Draft-State: encrypt=yes; _by-choice=yes;
+
+.. note::
+
+    It would be desirable to store this header in the encrypted
+    payload.  This is missing here because of technical limitations in
+    some MUAs.
```
dkg: "this is missing here" is confusing.  do you mean that a requirement is missing? or that we are demanding that it not be stored there for interoperability reasons?  we need to be more explicit about the requirements, i think.

doc/level1.rst
```diff
@@ -710,6 +701,79 @@ way:
...
+Autocrypt Gossip headers for drafts
++++++++++++++++++++++++++++++++++++
+
+When opening a draft that indicates it should be encrypted when sent
+via the ``Autocrypt-Draft-State`` header, the resuming MUA might be
+lacking keys for recipients that were originally available when the
+draft was stored.
+
+To solve this problem, a MUA MAY make use of the `Autocrypt
+Gossip<autocryt-gossip>` mechanism to include the keys of intended
```
dkg: link is misspelled.  can just be written as:

    :ref:`key-gossip`

no need to include link text explicitly, it'll get picked up from the markup.

doc/level1.rst
```diff
@@ -710,6 +701,79 @@ way:
...
+recipients in the draft message. To do so, it simply places one
+``Autocrypt-Gossip`` header per recipient in the MIME header of the
+encrypted payload of the draft.
+
+Conversely, when opening a draft, a MUA MAY import keys from
+``Autocrypt-Gossip`` headers that are present in the MIME headers of
+the encrypted payload.  These headers should update the ``gossip_key``
+and ``gossip_timestamp`` fields of the relevant peers `as
+usual<autocryt-gossip-update>`.
```
dkg: link is spelled wrong, and should be `key-gossip-update` to match the above change.

doc/level1.rst
Valodim: An attacker doesn't know the plaintext, so they can at best blindly inject drafts into the user's mailbox - which they can already do in plaintext. And considering the data is encrypted to the user's key, does the signed document for publication part really matter? I tried adding that part in some phrasings just now, but I felt like it made the pargraph unnecessarily wordy.

Valodim:
I don't know my original considerations about moving the section, but looking at it now: The section was effectively promoted from a subsection of "Message encryption" next to "Cleartext replies to encrypted messages", to its own section between Peer State Management and Managing Accounts. That still feels right to me, it is neither part of public key nor secret key management and provides enough value for coordinating co-MUA interoperability to stand on its own.

I updated some things from your other points, please re-review.

Valodim:
@dkg can I get a re-review on this one? :)

Valodim:
Ah, about the non-critical. We talked about this shortly after the comment was made which is why I forgot to reply :) summarizing here: The reason for using non-critical attributes is that all attributes besides "encrypt" are not strictly necessary for the interpretation of the header. A MUA needs to be able to parse the "encrypt=yes" to know the state of the encrypt button. Everything else is more granular information about that state.

Critical packets have unfortunately been a consistent source of confusion. If anyone is still confused, here's an example:

`encrypt=yes _is-reply-to-encrypted=yes` = this mail should be encrypted (the reason is that it's a reply to an encrypted message, but if you don't know what that means, don't worry about it)

`encrypt=yes is-reply-to-encrypted=yes` = this mail should be encrypted because it's a reply to an encrypted message. if you don't know what that means, disregard this entirely.

doc/level1.rst
```diff
@@ -722,6 +714,83 @@ way:
 4. Set ``peers[gossip-addr].gossip_key`` to the value of the
    ``keydata`` attribute.
 
+Message Drafts
+-------------
```
dkg: underscore too short: needs another - to match the title

doc/level1.rst
```diff
@@ -722,6 +714,83 @@ way:
...
+    Autocrypt-Draft-State: encrypt=yes; _by-choice=yes;
+
+.. note::
+
+    To minimize the amount of exposed metadata, it would be desirable
+    to store this header in the encrypted payload of the message,
+    rather than its outer envelope.  However, this causes technical
+    difficulties in some MUAs, which is why it is left out here to
+    optimize for interoperability.
+
```
dkg: This is missing text about stripping Autocrypt-Draft-State from the message before it is sent.

I think we should require MUAs to not send messages with this header.

doc/level1.rst
```diff
@@ -722,6 +714,83 @@ way:
...
+mechanism to include the keys of intended recipients in the draft
+message. To do so, it simply places one ``Autocrypt-Gossip`` header
+per recipient in the MIME header of the encrypted payload of the
+draft.
+
+Conversely, when opening a draft, a MUA MAY import keys from
+``Autocrypt-Gossip`` headers that are present in the MIME headers of
+the encrypted payload.  These headers should update the ``gossip_key``
+and ``gossip_timestamp`` fields of the relevant peers `as
+usual<key-gossip-update>`.
```
dkg: Are there security considerations about importing gossip from drafts?  I don't think i've thought it through as well as i should have.  it'd be nice to have some "here's what you need to think about when implementing this" note, if there are any gotchas.

Valodim: Considering we usually load these headers from wholly uncontrolled incoming mails, we're off no worse here even if the draft storage can be tampered with.

Valodim:
Added an example. This has gotten wordy for sure, we might want to consider splitting off sub-specs.

dkg:
it is wordy, but the functionality seems plausible and i think the definition is fine.  It is a significant piece of functionality, even if it won't be used super frequently.

PR #377, opened 2018-04-27
add FAQ entry about verification

Valodim:
This is an actual FAQ that keeps coming up, even in the iX printed article. We also need something citable for the wikipedia page.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -1,6 +1,29 @@
 Frequently Asked Questions about Autocrypt
 ==========================================
 
+Why is there no verification mechanism?
+---
+
+There are a couple of reasons for holding off on this for now:
+
+* Most OpenPGP implementations support verification of keys by asking
+  the user to compare fingerprints, so in practice there is already
+  a way to do this in most clients.
+* Conversely, since there is established tradition, there will likely
+  be pushback against whatever mechanism we introduce.
+* Verification requires user interaction, and introduces a lot of
+  friction. But even without verification, encrypted messages leak
+  much less information than plaintext. For that reason, getting key
+  management right is currently the more pressing issue.
+* If a user verifies a key on one device, we have no way of
+  synchronizing this to other devices. This is a difficult problem to
+  solve in e-mail, and leads to a shoddy user experience at best.
+
+That said, we do plan to introduce a simple but effective verification
+mechanism at some point. It will likely be modelled after Signal and
+WhatsApp's safety numbers, a solid workflow that many users are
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -1,6 +1,29 @@
...
+  the user to compare fingerprints, so in practice there is already
+  a way to do this in most clients.
+* Conversely, since there is established tradition, there will likely
+  be pushback against whatever mechanism we introduce.
+* Verification requires user interaction, and introduces a lot of
+  friction. But even without verification, encrypted messages leak
+  much less information than plaintext. For that reason, getting key
+  management right is currently the more pressing issue.
+* If a user verifies a key on one device, we have no way of
+  synchronizing this to other devices. This is a difficult problem to
```
hpk42: "... to other devices, yet".

doc/faq.rst
```diff
@@ -1,6 +1,29 @@
...
+  solve in e-mail, and leads to a shoddy user experience at best.
+
+That said, we do plan to introduce a simple but effective verification
+mechanism at some point. It will likely be modelled after Signal and
+WhatsApp's safety numbers, a solid workflow that many users are
+already familiar with.
```
hpk42: Are really that many users familiar with it?  I suggest it's possible that e.g. Whatsapp fingerprint validation work flows are about as popular/used as classic PGP/keyserver-mediated encryption is in e-mail.  I suggest to rather write more oppenly, e.g.: 

That said, many of us want to introduce guidance on verifications in future Autocrypt specifications. Stay tuned. 

FWIW I am all for opening this discussion and i hope to co-publish a paper with more information, discussion on active attacks against Autocrypt soon.  This paper does not prescribe what autocrypt is going to do, but is meant as a contribution and a widening of the discussion around preventing active attacks.

doc/faq.rst
```diff
@@ -1,6 +1,29 @@
 Frequently Asked Questions about Autocrypt
 ==========================================
 
+Why is there no verification mechanism?
```
hpk42: better "key verification" let's not be too implicit ... Saying "key verification" already hints at a conflict with Autocrypt's stated design philosophy, btw: "Don't talk to users about keys, ever". Just saying :)

Valodim: is this actually stated somewhere as a design philosophy? we can get around a lot of "key" language, but I don't really think the term itself is the core of the problem - explaining asymmetric crypto is.

anyways, agreed about the not too implicit :+1:

PR #379, opened 2018-05-03
Clarifications about user ids in key gossip

Valodim:
I found that the gossip section doesn't make it explicit that only the relevant user ids should be sent in gossip headers. This is only a practical consideration: in a pure autocrypt client it wouldn't come up, but actual implementations tend to merge keys that have the same fingerprints. For that reason, it should be ensured that user ids other than the relevant one (from the `addr` attribute) are stripped.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -678,9 +678,9 @@ each of which:
   for how this key is selected.
 
 - If a key has multiple user ids, only one SHOULD be contained in
-  ``keydata``.  If present, that user id SHOULD match the ``addr``
-  attribute.  This is only relevant for keys which came from or were
-  merged with data from external sources.
+  ``keydata``.  This user id SHOULD be picked to match the ``addr``
+  attribute, if possible.  This is only relevant for keys which came
+  from or were merged with data from external sources.
 
 - SHOULD NOT include a ``prefer-encrypt`` attribute.
 
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -677,9 +677,10 @@ each of which:
   referenced by ``addr``. See also :ref:`preliminary recommendation`
   for how this key is selected.
 
-- SHOULD contain only a single user id in the key material from the
-  ``keydata`` attribute, which matches the ``addr`` attribute. This
-  avoids leaking unrelated identities between recipients.
+- If a key has multiple user ids, only one SHOULD be contained in
+  ``keydata``.  If present, that user id SHOULD match the ``addr``
+  attribute.  This is only relevant for keys which came from or were
+  merged with data from external sources.
 
 - SHOULD NOT include a ``prefer-encrypt`` attribute.
 
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -677,6 +677,10 @@ each of which:
   referenced by ``addr``. See also :ref:`preliminary recommendation`
   for how this key is selected.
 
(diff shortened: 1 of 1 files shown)


doc/level1.rst
```diff
@@ -677,6 +677,10 @@ each of which:
   referenced by ``addr``. See also :ref:`preliminary recommendation`
   for how this key is selected.
 
+- SHOULD contain only a single user id in the key material from the
+  ``keydata`` attribute, which matches the ``addr`` attribute. This
+  avoids leaking unrelated identities between recipients.
```
hpk42: huh? why suddenly tie the user id to the addr? we say in the definition of the header format: "The content of the user id packet is only decorative.".

r10s: i think the pr belongs to the irc chat from yesterday. 
imu, it's not about to create a user id packet _belonging_ to addr; it's more about stripping all user id packets _not_ belonging to addr (eg. if the key is no autocrypt key and has legacy identities, ancient data and so on).

hpk42: then i think this third "SHOULD" point should just be a clarification to the second MUST eg:

 - MUST include the ``keydata`` attribute which MUST contain the
   same public key which is used to encrypt the mail to the recipient
   referenced by ``addr``.     See also  :ref:`preliminary recommendation`
   for how this key is selected. Note that for non-Autocrypt keys, only the one user id 
   which relates to ``addr`` SHOULD be contained in  ``keydata``.

dkg:
We have at least one implementation (muacrypt, iirc) that generates user IDs that don't match the actual e-mail address.  if i get a key from them, then i won't be able to meet this SHOULD.  as RFC 2119 says, SHOULD means:

> there may exist valid reasons in particular circumstances to ignore a particular item

Is not having a matching User ID the only reason any MUA should fail to meet this recommendation?  if so, we should spell that out here.  If we think there are other reasons, we should probably try to enumerate them.

doc/level1.rst
Valodim: You're right, I had totally forgotten about the decorative part. It's indeed only about stripping out packets that aren't relevant, if more than one is available. I worded this differently.

I added this I think keeping this as a separate point is reasonable since it's only relevant if the key is somehow merged or comes from an external source. I also added a more explicit note about that

Valodim:
I clarified that this is only relevant if there is more than one uid, and changed the wording so that only one should be sent, and that should be the one from addr if available.

doc/level1.rst
```diff
@@ -677,6 +677,11 @@ each of which:
   referenced by ``addr``. See also :ref:`preliminary recommendation`
   for how this key is selected.
 
+- If a key has multiple user ids, only one SHOULD be contained in
+  ``keydata``.  If present, that user id SHOULD match the ``addr``
+  attribute.  This is only relevant for keys which came from or were
```
hpk42: thanks. the "If present" confused me a little as i thought "why would a user id not be present?". 

Merging the first two sentences like this would unconfuse me and be shorter. 

"If a key has multiple user ids, only one SHOULD be contained in
 ``keydata`` and it SHOULD match the ``addr`` attribute. This is only ..."

Valodim: I see how that was confusing. I'm not too fond of this type of "SHOULD" in your suggestion though: It states a requirement that might not be possible to fulfill if a matching uid just isn't present.

I made another attempt

PR #383, opened 2018-06-02
Add terminology for key words used to indicate requirement levels

### LINKED ISSUE #381: Missing definition of keywords like "MUST", "MAY" etc.
In [Level 1 Spec](https://autocrypt.org/level1.html) I am missing a definition of the meaning of keywords like "MUST", "MAY" etc.

wodry:
fixes #381 
The update to RFC 2119 that states that RFC 2119 is only regarding uppercase key words, is [RFC 8174 "Ambiguity of Uppercase vs Lowercase in RFC 2119 Key Words"](https://tools.ietf.org/html/rfc8174)

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -59,6 +59,16 @@ both bugfix and backward-compatible feature releases.
    \sphinxtableofcontents
    \newpage
 
+Terminology
+-----------
+
+Key words to indicate requirement levels
+++++++++++++++++++++++++++++++++++++++++
+
+The definition of uppercase key words used in this document to indicate requirement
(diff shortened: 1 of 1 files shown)


hpk42:
lgtm -- is  it comon to say "key word" instead of "keyword"? FWIW my understanding/language sense if probably tainted by a german nouncoupling practise ;)

wodry:
I took the spelling over from thew two RFCs, I like keyword better, but have no real opinion.
Seems to be both possible
http://www.dictionary.com/browse/keyword
https://en.wikipedia.org/wiki/Keyword_(linguistics)   (where both spellings are mixed)

wodry:
Will make a new pull request.

PR #384, opened 2018-06-07
Add terminology for requirement level keywords

### LINKED ISSUE #381: Missing definition of keywords like "MUST", "MAY" etc.

wodry:
Fixes #381

Regarding the question of spelling "key word" or "keyword":

RFC 2119 explicitly recommnends how to reference to itself:
>Authors who follow these guidelines
>should incorporate this phrase near the beginning of their document:
>
>The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
>"SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL"
>in this document are to be interpreted as described in RFC 2119.

In the Autocrypt document itself I chose here "keyword", which seems to be more common English. Since RFC 2119 states that the given phrase only "should" (not "SHOULD" BTW ;) be incorporated, I guess it's ok to change "key words" in the RFC phrase to "keywords" to have a consistent spelling in Autocrypt documentation and in the same time comply to RFC 2119.

I would suggest to leave any further linguistic discussion open for following pull requests.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -59,6 +59,17 @@ both bugfix and backward-compatible feature releases.
    \sphinxtableofcontents
    \newpage
 
+Terminology
+-----------
+
+Keywords to indicate requirement levels
++++++++++++++++++++++++++++++++++++++++++
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in
+:rfc:`RFC 2119 "Key words for use in RFCs to Indicate Requirement Levels"<2119>`.
+
 Overview
 --------
 


wodry:
Help with LaTeX formating appreciated, if that should be the cause for the Travis failure.

wodry:
apt-get error in Travis because of several apt repository URLs connect timeout, e.g.
>Err:35 https://packagecloud.io/computology/apt-backport/ubuntu trusty InRelease
Failed to connect to packagecloud.io port 443: Connection timed out
Err:42 http://mirror.jmu.edu/pub/ubuntu trusty InRelease
Could not connect to mirror.jmu.edu:80 (134.126.36.183), connection timed out

Could You trigger Travis build again, maybe the connections work now again?

hpk42:
thanks, it's all passing.  i am all for merging. please someone who agrees do :)

Valodim:
I'm fine with the content of this PR, but we should really talk about versioning. Is this still version 1.0?

dkg:
RFC 8174 offers clarifying text to indicate that only the upper-case variants should be considered 2119ese:

> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
> NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
> "MAY", and "OPTIONAL" in this document are to be interpreted as
> described in BCP 14 [RFC2119] [RFC8174] when, and only when, they
> appear in all capitals, as shown here.

I'll update this PR and merge it unless someone objects.

I agree with @Valodim that we need to have a better versioning plan, but i don't think this change should block on that.

PR #386, opened 2018-07-09
Update events

dkg:


r10s:
looks good to me, merging before the future events become the past events :)

PR #391, opened 2018-09-10
Allow gossip header for reply-to address

azul:
In order to enable encrypted replies in the presence of a Reply-To header
we allow specifying Autocrypt-Gossip for the given address.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -672,35 +672,38 @@ attribute indicates the recipient address this header is valid for as
 usual, but may relate to any recipient in the ``To`` or ``Cc`` header.
 See example in :ref:`autocrypt-gossip-example`
 
-The ``Autocrypt-Gossip`` header can also be used
+The ``Autocrypt-Gossip`` header MAY also be used
 to include keys for the address specified in the ``Reply-To`` header.
+This allows replying encrypted even if the address differs from those in
+the ``From``, ``To``, and ``Cc`` headers.
 
 Key Gossip Injection in Outbound Messages
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-An Autocrypt MUA MAY include ``Autocrypt-Gossip`` headers in messages
-with more than one recipient. These headers MUST be placed in the root
-MIME part of the encrypted message payload. The encrypted payload in
-this case contains one Autocrypt-Gossip header for each recipient,
-each of which:
+An Autocrypt MUA MAY include ``Autocrypt-Gossip`` headers in messages.
+These headers MUST be placed in the root MIME part of the encrypted message payload.
+The encrypted payload in this case contains
+one Autocrypt-Gossip header for each address.
+Each header:
 
 - MUST include an ``addr`` attribute that matches one of the
-  recipients in the ``To``, ``Cc``, or ``Reply-To`` headers.
+  addresses in the ``To``, ``Cc``, or ``Reply-To`` headers.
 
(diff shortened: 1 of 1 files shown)


azul:
Thinking about this...
What happens if I receive an encrypted mail with a Reply-To header right now?
Will a reply create an unencrypted message to that address?
Is there a warning?

doc/level1.rst
```diff
@@ -682,12 +685,15 @@ this case contains one Autocrypt-Gossip header for each recipient,
...
+  recipients in the ``To``, ``Cc``, or ``Reply-To`` headers.
 
-- MUST include the ``keydata`` attribute which MUST contain the
+- MUST include the ``keydata`` attribute.
+  For ``To`` and ``Cc`` headers it MUST contain the
   same public key which is used to encrypt the mail to the recipient
   referenced by ``addr``. See also :ref:`preliminary recommendation`
   for how this key is selected.
+  For ``Reply-To`` headers it MUST contain the public key
+  which should be used to encrypt a reply.
```
hpk42: what if a sender has the keys for To/CC recipients (which it must in order to encrypt) but not for Reply-To?

Valodim: this is per individual header, so in that case there simply wouldn't be a header with an `addr` to match the `Reply-To`

dkg: I'm not convinced that MUST is particularly meaningful in this new sentence -- how could it be contradicted?  I'm also pedantically concerned that it seems to imply that any reply (even those that aren't addressed to the `Reply-To` address) should be encrypted to that key.

doc/level1.rst
```diff
@@ -682,12 +685,15 @@ this case contains one Autocrypt-Gossip header for each recipient,
 each of which:
 
 - MUST include an ``addr`` attribute that matches one of the
-  recipients in the ``To`` or ``Cc`` headers.
+  recipients in the ``To``, ``Cc``, or ``Reply-To`` headers.
```
dkg: I think this should change "recipients" to "addresses", since the address in a `Reply-To` header is not a recipient.

doc/level1.rst
```diff
@@ -672,6 +672,9 @@ attribute indicates the recipient address this header is valid for as
 usual, but may relate to any recipient in the ``To`` or ``Cc`` header.
 See example in :ref:`autocrypt-gossip-example`
 
+The ``Autocrypt-Gossip`` header can also be used
+to include keys for the address specified in the ``Reply-To`` header.
```
Valodim: It's pedantic, but we might want to mention that this is only useful in cases where that address differs from the ones in From/To/Cc

doc/level1.rst
```diff
@@ -710,8 +716,8 @@ Autocrypt peer state of the gossiped recipient identified by its
 way:
 
 1. If ``gossip-addr`` does not match any recipient in the mail's
-   ``To`` or ``Cc`` header, the update process terminates (i.e.,
-   header is ignored).
+   ``To``, ``Cc``, or ``Reply-To`` header,
```
Valodim: same as above, "recipient" should now be "address"

azul:
Updated the PR:
* talking about `addresses` rather than `recipients`
  because Reply-To headers contain an address but not a recipient.
* using SHOULD rather than MUST for picking the key for
  Reply-To gossip
* explaining Reply-To gossip makes sense if the address is not
  in To or Cc headers already.
* Dropped the requirement for the message to have multiple recipients.
  Gossip may make sense if only one recipient but also a Reply-To header
  are present.

PR #392, opened 2018-11-01
Add key id in example

oliverFU:
Hi,
I added the example autocrypt header (https://autocrypt.org/level1.html#e-mail-address-canonicalization) to my test suite. I think the key id is 71DBC5657FDE65A7 and this was helpful for testing. May other can verify the key id and other find this helpful, too.

Best 
Oliver


Valodim:
Looks good, thanks!

PR #393, opened 2018-11-09
Switch to Ed25519+Cv25519 in v1.1

Valodim:
As discussed at the meeting.

RSA remains a SHOULD. I put a remark about older versions into a note, that seemed like the appropriate place.


doc/level1.rst
```diff
@@ -297,10 +297,21 @@ concerns, the user id SHOULD NOT be an empty string.
 These packets MUST be assembled in binary format (not ASCII-armored),
 and then base64-encoded.
 
-A Level 1 MUA MUST be capable of processing and handling 2048-bit and
-3072-bit RSA public keys.  It MAY support other OpenPGP key formats
-found in an Autocrypt header (for example, by passing it agnostically
-to an OpenPGP backend for handling).
+A Level 1 MUA MUST be capable of processing and handling Ed25519
```
r10s: shouldn't this read "A Level 1.1 MUA ..."?

Valodim: Nope. This is still level 1, but version 1.1 of it.

r10s: yeah, as mentioned, i understand. but still find this confusing ;)
but it's fine for me :)

hpk42:
On Mon, Nov 12, 2018 at 07:16 +0000, Patrick Brunschwig wrote:
> pbrunschwig approved this pull request.
> 
> Fine with me. 
> 
> @r10s The Level is an overall concept, whereas the version is the revision of the document.

however, saying "Level 1.1 client" is easier and will be chosen often over
"a Level 1 client as specified with revision 1.1" or so ;)

doc/level1.rst
```diff
@@ -297,10 +297,21 @@ concerns, the user id SHOULD NOT be an empty string.
...
+both signatures and encryption.
+
+.. note::
+
+  To optimize for compatibility, earlier versions of this document
+  REQUIRED support for RSA, and recommended a 3072 bit RSA key
+  configuration.  Support for elliptic curve cryptography in deployed
+  OpenPGP implementations improved since then, and the switch to ECC
+  was made in version 1.1 (December 2018) due to message size
+  considerations.
```
hpk42: i remain fine with this, good note :+1:

dkg:
The typical way that i'd expect such a change to happen in a protocol doc with wider adoption is to require support for handling the new crypto primitives (when seen from communications peers) in version X+1, and then mandate generation of the new key types in version X+2.

Perhaps we're small enough and we have a wide enough canvas already that we can just jump ahead, though it makes me slightly nervous. [PGPy still doesn't support cv25519 and ed25519](https://github.com/SecurityInnovation/PGPy/issues/221), for instance, though perhaps its current stagnation is a warning about that library itself. :(

I'm open to the idea, though -- i definitely like the size improvements we're seeing with the ECC keys.

This change should not be merged without updates to the appendix, though -- the example keys and messages should all be re-created using 25519.

Valodim:
I updated the examples. I already double-checked myself, anyone care to double-double-check? Also added a script for generation. It works, but better not look at that too closely ;)

Valodim:
@oliverFU now :)

PR #397, opened 2019-01-16
note: MUAs MUST NOT generate headers > 10 KiB

azul:
In theory one can construct larger Autocrypt headers in a spec conform way
(for example by adding non critical attributes).

This is likely to cause other problems though.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -275,6 +275,22 @@ above rules, such headers will be judged invalid, and discarded by
 level 1 MUAs.  Such an update to the specification will also have
 to describe how an updated MUA will deal with multiple valid headers.
 
+A MUA MUST NOT send an Autocrypt header above 10 KiB in size
+(including the Autocrypt: prefix and folding whitespace).
+Larger headers are likely the result of unintentionally bloated OpenPGP data.
+To ensure a consistent failure mode across clients,
+MUAs SHOULD treat such headers as invalid.
+
+.. note::
+
+  - An Autocrypt header following this specification
+    will be up to 3 KiB in size.
+    Some users might import 4096-bit RSA keys,
+    which may result in an Autocrypt header of up to 5 KiB.
+  - The 10 KiB limit was chosen to accomodate for these use cases
+    but also avoid headers being truncated or emails being rejected
+    by MTAs with a 10 KiB header size limit.
+
 
 OpenPGP Based key data
 ~~~~~~~~~~~~~~~~~~~~~~


doc/level1.rst
```diff
@@ -302,6 +319,7 @@ A Level 1 MUA MUST be capable of processing and handling 2048-bit and
 found in an Autocrypt header (for example, by passing it agnostically
 to an OpenPGP backend for handling).
 
+
```
dkg: why is this extra whitespace here?

doc/level1.rst
```diff
@@ -275,6 +275,23 @@ above rules, such headers will be judged invalid, and discarded by
 level 1 MUAs.  Such an update to the specification will also have
 to describe how an updated MUA will deal with multiple valid headers.
 
+A MUA MUST NOT send Autocrypt headers above 10 KiB in size
```
dkg: In a spec, I tend to prefer to leave things in the singular, until there is a clear reason to put it in the plural.  In the line above, it could be (mis)read as suggesting that there are multiple autocrypt headers that are in aggregate bounded by 10KiB.  can we re-frame it in singular instead? (e.g. "An MUA MUST NOT send an Autocrypt or Autocrypt-Gossip header greater than 10KiB in size…")

doc/level1.rst
```diff
@@ -275,6 +275,23 @@ above rules, such headers will be judged invalid, and discarded by
...
+(including the Autocrypt: prefix and folding whitespace).
+Conversely, MUAs SHOULD not process Autocrypt headers
```
dkg: Again, i would prefer singular here.  and what should the MUA do if it finds an Autocrypt header that is > 10KiB in size?  should it treat the header as non-existent?  as an invalid or mal-formed Autocrypt header?  Those might be subtly different, and we should give clear guidance.

doc/level1.rst
```diff
@@ -275,6 +275,23 @@ above rules, such headers will be judged invalid, and discarded by
...
+larger than 10 KiB in size.
+
+.. note::
+
+  - An Autocrypt header following this specification
+    will be up to 3 KiB large.
```
dkg: we say "in size" everywhere else.  maybe s/large/in size/ here too?

doc/level1.rst
```diff
@@ -275,6 +275,23 @@ above rules, such headers will be judged invalid, and discarded by
...
+    Some users import 4096-bit RSA key configurations,
```
dkg: itym "Some users *might* import" -- and i'm not sure whether "configurations" is the right term.  can we just say "4096-bit RSA keys"

doc/level1.rst
```diff
@@ -275,6 +275,23 @@ above rules, such headers will be judged invalid, and discarded by
...
+    which may result in Autocrypt headers of up to 5 KiB.
```
dkg: a 4096-bit RSA primary key with a single user id and two 4096-bit subkeys (one for signing, one for encryption), base64-encoded, comes to about 5256 octets plus the length of the base64-encoded user ID.  That's slightly more than 5KiB, and it doesn't even include the header name or other autocrypt attributes.  maybe we want to say "about 5KiB" or "less than 6KiB"?

doc/level1.rst
```diff
@@ -275,6 +275,23 @@ above rules, such headers will be judged invalid, and discarded by
...
+  - The 10 KiB limit was chosen to accomodate for these use cases
+    but also avoid headers being truncated or emails being rejected
+    by MTAs with a 10 KiB header size limit.
+    In addition it makes header sizes that are indicative of bugs
+    obvious to implementors.
```
dkg: This last sentence is a great point, but i'd phrase it the other way around.  It makes it easy for implementors of receiving MUAs to reject buggy headers that happen to be oversize.

doc/level1.rst
azul: I checked with my own AC header - pasted it to a file an ran wc on it - which resulted in approx 3200 chars.
I'll try to figure out what's going on there.

azul:
```
$ ls -l azul.b64 
-rw-r--r-- 1 azul azul 3054 Jan 17 08:42 azul.b64
$ wc azul.b64 
  42   42 3054 azul.b64
$ base64 -d azul.b64 | gpg --list-packets
# off=0 ctb=99 tag=6 hlen=3 plen=525
:public key packet:
	version 4, algo 1, created 1324285022, expires 0
	pkey[0]: [4096 bits]
	pkey[1]: [17 bits]
	keyid: 786085AC20D3457D
# off=528 ctb=b4 tag=13 hlen=2 plen=23
:user ID packet: "Azul. <azul@riseup.net>"
# off=553 ctb=89 tag=2 hlen=3 plen=599
:signature packet: algo 1, keyid 786085AC20D3457D
	version 4, created 1538491089, md5len 0, sigclass 0x13
	digest algo 10, begin of digest 16 08
	hashed subpkt 27 len 1 (key flags: 23)
	hashed subpkt 30 len 1 (features: 01)
	hashed subpkt 23 len 1 (keyserver preferences: 80)
	hashed subpkt 25 len 1 (primary user ID)
	hashed subpkt 11 len 4 (pref-sym-algos: 9 8 7 3)
	hashed subpkt 21 len 4 (pref-hash-algos: 10 9 8 11)
	hashed subpkt 22 len 4 (pref-zip-algos: 2 3 1 0)
	hashed subpkt 33 len 21 (issuer fpr v4 BB70B0160449010A00AF49E8786085AC20D3457D)
	hashed subpkt 2 len 4 (sig created 2018-10-02)
	hashed subpkt 9 len 4 (key expires after 7y289d5h41m)
	subpkt 16 len 8 (issuer key ID 786085AC20D3457D)
	data: [4095 bits]
# off=1155 ctb=b9 tag=14 hlen=3 plen=525
:public sub key packet:
	version 4, algo 1, created 1509612037, expires 0
	pkey[0]: [4096 bits]
	pkey[1]: [17 bits]
	keyid: 566E95B3ECCE1EE8
# off=1683 ctb=89 tag=2 hlen=3 plen=572
:signature packet: algo 1, keyid 786085AC20D3457D
	version 4, created 1538491141, md5len 0, sigclass 0x18
	digest algo 10, begin of digest e2 e1
	hashed subpkt 27 len 1 (key flags: 0C)
	hashed subpkt 33 len 21 (issuer fpr v4 BB70B0160449010A00AF49E8786085AC20D3457D)
	hashed subpkt 2 len 4 (sig created 2018-10-02)
	hashed subpkt 9 len 4 (key expires after 1y334d5h58m)
	subpkt 16 len 8 (issuer key ID 786085AC20D3457D)
	data: [4091 bits]
```

azul:
I tried to address the comments of @dkg and marked those requests as resolved that seemed clear for me.

I'm curious to see what will be the result of the size discussion. I pasted my empirical results.
I will not be able to follow up on this for the next month. I'd be happy if someone feels like picking this up and finalizes it in the meantime.

doc/level1.rst
Valodim: Your key doesn't have independent signing and certification keys. This is what Autocrypt recommends, but for an "upper bound" we'll have to consider the 3 x RSA 4096 configuration as dkg says

doc/level1.rst
dkg: works for me :)

doc/level1.rst
dkg: yep, i think @Valodim has identified the disparity -- it's not so much *separate* as "a signing-capable subkey" (since the primary could be signing+certification capable, and a signing-capable subkey could also be present).

doc/level1.rst
Valodim: This was also neither changed nor commented, but marked as resolved?

doc/level1.rst
Valodim: The intent of my phrasing was that a MUA that sends faulty autocrypt headers will simply fail to work if the oversized headers are rejected. This would surely have caught issues like the ones Enigmail initially had.

I'm unsure what "makes it easier" would even mean in this context for a receiving MUA?

(this was also marked as resolved but not changed nor commented?)

doc/level1.rst
Valodim: similarly resolved but not changed nor commented? @azul I'm confused by your understanding of "resolve" :)

doc/level1.rst
azul: Sorry... I did not notice that the push of the amended commit failed.
Now should be fine.

azul:
oh my... made the changes locally, marked the comments as resolved. Then the push failed and i went afk for a month.
Sorry for the mess. Thanks for pointing it out @Valodim.
Hope things are better now.

doc/level1.rst
azul: sorry... had forgotten to push. Should now actually be resolved.

doc/level1.rst
azul: sorry... had forgotten to push. Should now actually be resolved.

doc/level1.rst
azul: Here's my latest take on the last sentence:

> Sending an oversized autocrypt header will result in consistent
    behaviour accross various clients.
    This way the issue can be detected more easily.

@Valodim does it capture your thought?
@dkg works for you?

doc/level1.rst
azul: @dkg, leaving out the gossip part for now.

doc/level1.rst
```diff
@@ -275,6 +275,24 @@ above rules, such headers will be judged invalid, and discarded by
...
+
+  - An Autocrypt header following this specification
+    will be up to 3 KiB in size.
+    Some users might import 4096-bit RSA keys,
+    which may result in an Autocrypt header of up to 5 KiB.
+  - The 10 KiB limit was chosen to accomodate for these use cases
+    but also avoid headers being truncated or emails being rejected
+    by MTAs with a 10 KiB header size limit.
+  - Sending an oversized autocrypt header will result in consistent
+    behaviour accross various clients.
```
Valodim: This is phrased in an oddly positive way now, as if sending an oversized header was intended behavior ("will result in consistent behavior"). I suggest we drop this line and clarify it in the non-note paragraph above:

```
Headers larger than the 10kb are likely the result of unintentionally bloated OpenPGP data. To ensure a consistent failure mode across clients, MUAs SHOULD treat such headers as invalid.
```

Valodim: This goes back to a typical postel discussion: If there is a usable key in the keydata, why disregard it? Wouldn't it better to accept it and not get in the user's way?

I would say: To an extent. If the key is 8kb in size (perhaps due to a third party signature), it's fine, don't bother. If it's 1MB in size (as an unminimized key coming from a user's keyring might well be), it would certainly be better to fail hard enough that users notice. Otherwise it might mask nasty bugs, as happened intially with Enigmail.

azul: @Valodim I used a slight variation of the phrase you proposed. I think it makes the reasoning easier to understand.

PR #398, opened 2019-01-21
Update dev-status.rst

hpk42:
muacrypt since the last public releases supports recommend-algo, encryption and the setup process guidance, and gossip.


lefherz:
cool!

PR #399, opened 2019-02-03
Fix attribution links in the code of conduct

blinry:
The links were previously in Markdown format, which did funny things to
https://autocrypt.org/code_of_conduct.html.


hpk42:
thanks!

PR #405, opened 2019-04-22
simple cleanup

dkg:
this is simple bookkeeping stuff, and shouldn't be at all controversial


hpk42:
"Your attention to detail is appreciated" :)

PR #413, opened 2019-10-15
allow AES-128 or AES-256 for autocrypt setup message

### LINKED ISSUE #390: Autocrypt Setup Message Example uses AES-256, not AES-128
The [section on the ASM](https://autocrypt.org/level1.html#autocrypt-setup-message) says:

```
The symmetric encryption algorithm used MUST be AES-128.
```

But, [the example message ](https://autocrypt.org/level1.html#setup-message-example) is encrypted using AES-256.

Valodim:
Fixes #390


BjarniRunar:
I just read the linked issue, seems the goal is to just align the spec with the reality of what tools are doing. That seems like a fine reason to me.

dkg:
thanks for doing this.

PR #417, opened 2019-10-24
Some CSS to make the dev-status table more legible

BjarniRunar:
This rotates the column headings 90°, making the colunms narrower, so things actually fit on smaller screens. Also adds subtle striping to the table to make it easier to scan.

image


lefherz:
Wow, this is awesome. I always hated that scroll bar, but didn't think of that solution.

BjarniRunar:
I'll do the rebase dance :-)

dkg:
fwiw, i think an angled rotation of less than 90°, and in the opposite direction would be better.

see the examples in https://css-tricks.com/rotated-table-column-headers/

i'll supply a patch if @BjarniRunar doesn't want to do it :)

dkg:
interesting, Chromium 76.0.3809.100-1 doesn't appear to support the `writing-mode` property, but Firefox 70.0 does. (this is in contradiction to [the MDN docs for `writing-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode)

dkg:
I did #421 based on this idea (and incorporating bjarni's initial work).  Please consider merging that one instead.

dkg:
I think this has been superceded by #421 , which is now merged.  Thanks for the inspiration, @BjarniRunar !

PR #420, opened 2019-10-31
Using gossip_key should result in an ui-recommendation?

user9209:
> If public_key is null, then set target-keys[to-addr] to gossip_key and set
    preliminary-recommendation to discourage and skip to the Deciding to Encrypt by Default
    (page 9).

If I understand it correctly, the use of gossip_key ends as a `preliminary-recommendation`, which is later never used, due to the fact that `Deciding to Encrypt by Default` is skipped.

So shouldn't the result of using the gossip_key be an ui-recommendation?

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -567,7 +567,7 @@ __________________________
 
 If ``public_key`` is ``null``, then set
 ``target-keys[to-addr]`` to ``gossip_key`` and set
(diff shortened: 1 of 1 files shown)


hpk42:
it says "skip to" not "skip".  A better phrase might be "continue with" or so.

user9209:
> it says "skip to" not "skip".

Ah ok, my mistake.
I close this pull request now.

PR #421, opened 2019-10-31
readable angled column headers on dev-status table.

dkg:
This is based on (and intended to supercede) #417

the dev-status table should look like this in browsers that support these transforms:

table


dkg:
I've just pushed one additional commit that makes it so that the header row and the MUA/Project column both are "sticky", so that readers who do need to scroll can always see the labels/headers.

dkg:
OK, there were some differences between sphinx 2.2.1 (used by the travis build) and 1.8.5 (in debian unstable, what i'm using), but i think i've accomodated them both in the current series.

hpk42:
looks fine to me. thanks for doing this!

r10s:
yeah, looks great! happy to merge that!

PR #422, opened 2019-11-04
Add reference to Protected Headers for Cryptographic E-mail

dkg:
Several of us contributed to this work, documenting what existing
implementations do.  Referring to it directly here will help others
find the reference.

Signed-off-by: Daniel Kahn Gillmor <dkg@fifthhorseman.net>


hpk42:
looks good -- i think we should start doing a changelog, at least in the core repo, but maybe better in the level1 doc.

lefherz:
Hey, what's the state of this? The draft expired :/

hpk42:
there is a new draft https://www.ietf.org/archive/id/draft-ietf-lamps-header-protection-18.html -- we can close this issue here.

PR #429, opened 2020-01-19
Ignore all automatically replied messages

link2xt:
multipart/report messages are not required to be placed at the root of
MIME message. RFC 8098 specifically allows it for Message Disposition
Notification. For example, reports may be combined into multipart/parallel
or multipart/mixed.

If these messages are properly marked as auto-replied, they can still
be ignored without parsing them recursively.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -412,9 +412,13 @@ header, by an MUA at receive or display time.
 Messages SHOULD be ignored (i.e., ``peers[from-addr]`` SHOULD NOT be
 updated) in the following cases:
 
+  - The message is explicitly marked as auto-replied with an
+    `Auto-Submitted: auto-replied` header (:rfc:`3834`).
+
   - The content-type is ``multipart/report``. In this case, it can be assumed
-    the message was auto-generated. This avoids triggering a ``reset``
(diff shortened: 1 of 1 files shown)


link2xt:
It might also make sense not to ignore them if they contain an `Autocrypt` header, because some clients may choose to send `Autocrypt` header in message disposition notifications, but it is a separate proposal.

doc/level1.rst
```diff
@@ -412,9 +412,13 @@ header, by an MUA at receive or display time.
 Messages SHOULD be ignored (i.e., ``peers[from-addr]`` SHOULD NOT be
 updated) in the following cases:
 
+  - The message is explicitly marked as auto-replied with an
+    `Auto-Submitted: auto-replied` header (:rfc:`3834`).
+
```
hpk42: Hum, is it it really right to completely ignore thos?  Delta Chat bots are going to use these headers at some point but still send Autocrypt headers.  Maybe we just need to say that with the Auto-Submitted header we leave the peerstate untouched if there is no Autocrypt header?  If an Autocrypt header is there, it's fine to interpret and update the peer-state i'd think.

hpk42:
closed for now, let's discuss how/if-reopening.

PR #432, opened 2020-04-10
Update references to py-autocrypt to rebranded name muacrypt

wodry:
py-autocrypt has been rebranded to muacrypt (see https://github.com/hpk42/muacrypt/issues/34)


hpk42:
thanks @wodry for taking care!

PR #434, opened 2020-06-22
Typo fix: "an MUA" -> "a MUA"

wodry:
See [When to Say *a* or *an*](https://www.englishclub.com/pronunciation/a-an.htm)


wodry:
No, I see "an MUA" is correct, because "M" is pronounced as "emm" which is a vocal sound.
(see "an FBI agent" as example in the link accordingly), I initially thought of pronouncing "myuua" or speaking it out "mail user agent", but when speaking M. U. A. "an" is correct.

hpk42:
interesting -- didn't know that a/an is based on vocal sounds.

PR #438, opened 2021-05-08
Fix broken link in doc/level1.rst

seifferth:
The link to <https://autocrypt.org/pgp-users> leads nowhere. I am
not sure if this page simply does not exist anymore or if it is not
implemented yet. In any case, the substituted link leads somewhere
at least.


lefherz:
Thanks a lot! sorry that this was overlooked for so long.

PR #441, opened 2022-01-10
Add Ltt.rs to dev-status page

iNPUTmice:
Ltt.rs 0.3.0 has native support for Autocrypt Level 1 (except Gossip headers)


lefherz:
cool! do you plan to add gossip headers in the future?

iNPUTmice:
> cool! do you plan to add gossip headers in the future?

Yes. Very soon.

The Autocrypt support was funded by the EU through NGIPointer and I was struggling to make my milestones review so anything that wasn’t explicitly mentioned as a requirement got pushed back to after. Gossip support being among that. However all the internal data structures and the recommendation engine already have support for Gossip. The headers just aren’t being written nor parsed. So it should actually be fairly easy to add once the pressure of making the milestone review is off.

lefherz:
Cool! Looking forward to exchange mails with ltt.rs-users :)

PR #451, opened 2024-06-17
Add Thunderbird native Autocrypt support to development status table 

missytake:
@kaie is this more or less correct?

I set `peer state` to unsupported because Thunderbird doesn't track the timestamp when a key was last seen, as far as I could tell from the UI. And pretty sure that it doesn't treat the user ID as decorative, e.g. the key of @hpk42 always gives me trouble because of this^^

The `gossip` column wasn't clearly defined, and Thunderbird only send `Autocrypt-Gossip` headers but doesn't seem to update peer state from received `Autocrypt-Gossip` headers. I first contemplated making this more clear in the column, but then again I don't know how it is for most of these implementations (though I'm pretty sure that if they have a checkmark for `gossip` they at least send out the header), so I was not comfortable with specifying that they all import keys from gossip headers. Maybe we should create another column for that at some point.

### DIFF
--- a/doc/dev-status.rst
+++ b/doc/dev-status.rst
@@ -93,8 +93,6 @@ _ ``gossip``: sends out Autocrypt-Gossip headers
 .. _delta.chat: https://delta.chat/
 .. _thunderbird: https://www.thunderbird.net/
 .. _K-9 Mail: https://k9mail.github.io/
-.. _Autocrypt-Thunderbird: https://addons.thunderbird.net/en-US/thunderbird/addon/autocrypt/
-.. _Enigmail: https://www.enigmail.net/
 .. _Mutt: http://mutt.org/
 .. _Neomutt: https://neomutt.org/
 .. _`Letterbox`: https://letterbox-app.org/
--- a/doc/dev-status.rst
+++ b/doc/dev-status.rst
@@ -1,7 +1,7 @@
 Autocrypt-capable MUAs level 1 implementation status
 ====================================================
 
-Last updated: ``2024-06-10``
+Last updated: ``2024-06-17``
 
 Note that the below table is not complete and not up-to-date.
 Many more mail user agents support at least a subset of Autocrypt.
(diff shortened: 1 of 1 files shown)


kaie:
Peer status is unsupported, correct.

Thunderbird uses autocrypt headers for consuming potential correspondent's public key, only, and for distributing public keys.

I'm not fully clear what "used ID is decorative" means, so let me explain how Thunderbird interprets the user ID.
Thunderbird requires that an OpenPGP key contains a user ID in order to use a key. It will only consider to use a key for an email address, if a user ID is found with the matching email address. The email address value in the Autocrypt header is used only as a first check to learn for which email address the key is intended. (The only way in Thunderbird to use a public key with a non-matching email address is by manually defining Alias rules.)

Thunderbird versions 115.7.0 and later both distribute and consume the keys present in the Gossip headers.
However, it doesn't immediately import them. Thunderbird keeps them in a cache of key candidates. Only when the user attempts to send an email to a correspondent for whom no accepted key is available yet, and the user decides fix that, then those candidate keys will be offered to the user.

missytake:
> I'm not fully clear what "user ID is decorative" means, so let me explain how Thunderbird interprets the user ID. Thunderbird requires that an OpenPGP key contains a user ID in order to use a key. It will only consider to use a key for an email address, if a user ID is found with the matching email address. The email address value in the Autocrypt header is used only as a first check to learn for which email address the key is intended. (The only way in Thunderbird to use a public key with a non-matching email address is by manually defining Alias rules.) 

yep, I noticed this behavior. The Autocrypt spec mentions in https://docs.autocrypt.org/level1.html#openpgp-based-key-data that the user ID is only decorative, I'm interpreting that as "the user ID should be ignored, only the `addr` value in the `Autocrypt` header matters". As Thunderbird treats an Autocrypt-transmitted key with a different user ID as unusable, it isn't interoperable with some usages of Autocrypt-transferred openPGP keys. But apart from @hpk42's key I have never encountered the problem.

> Thunderbird versions 115.7.0 and later both distribute and consume the keys present in the Gossip headers. However, it doesn't immediately import them into the permanent keyring. Thunderbird keeps them in a cache of key candidates, and thus treats them equally as other keys found elsewhere (e.g. from regular email attachments). Only when the user attempts to send an email to a correspondent for whom no accepted key is available yet, and the user decides fix that, then those candidate keys will be offered to the user.

Ah, this makes a lot of sense in Thunderbird's model :)

PR #454, opened 2024-11-17
Add Evolution to dev-status

adaschma:
Last year, GNOME's default email client Evolution started advertising the gpg-key via autocrypt.[1]

[1] https://gitlab.gnome.org/GNOME/evolution/-/issues/375


missytake (approved):
Thanks, great to hear!

PR #455, opened 2025-04-06
add 'Cold E-Mail' section and subsection mailto

iNPUTmice:
Current status of what I have for the mailto URI params.

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -1156,7 +1156,7 @@ The following query parameter are used by Autocrypt:
  * `autocrypt-keydata`: The Base64 representation of the binary :rfc:`OpenPGP "Transferable Public Key"<4880#section-11.1>` of the address the mailto URI is referring to, as specified for the Autocrypt header without newlines and with any padding characters percent-encoded.
  * `autocrypt-prefer-encrypt`: The prefer-encrypt value (either nopreference or mutual) of the address, as specified for the Autocrypt header. 
 
-The values derived from the mailto URI MUST not modify the existing Autocrypt Peer State for the given address. If the local Autocrypt Peer State has a key for the given address and it’s `preliminary-recommendation` is calculated to be `available` that key MUST take precedent over the one supplied through the `mailto` URI. The values MAY only be used for composing the e-mail that was invoked by the mailto URI.
+The values derived from the mailto URI MUST not modify the existing Autocrypt Peer State for the given address. The values MAY only be used for composing the e-mail that was invoked by the mailto URI. If the local Autocrypt Peer State has a key for the given address and it’s `preliminary-recommendation` is calculated to be `available` (meaning not outdated and not received through gossip), that key MUST take precedent over the one supplied through the `mailto` URI.
 
 VCard
 +++++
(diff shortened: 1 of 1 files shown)


missytake (commented):
Thanks for driving this forward :)

When discussing the vCard spec, we thought about moving it to a new document, linked on the side - but the more I think about it, it really fits under the Cold E-Mail section. @twiss what do you think?

In any case I would leave out vCards from this PR and add it in a subsequent PR.

I suggest to also add a column to the table on https://docs.autocrypt.org/dev-status.html, headed "mailto: URI". "vCards" and "Export Contacts" would be additional column headings I guess.

doc/level1.rst
```diff
@@ -1142,6 +1142,27 @@ looking for a Setup Message again.  A more sophisticated
...
+
+mailto: URI
++++++++++++
+
+The following query parameter are used by Autocrypt:
+
+ * `autocrypt-keydata`: The Base64 representation of the binary :rfc:`OpenPGP "Transferable Public Key"<4880#section-11.1>` of the address the mailto URI is referring to, as specified for the Autocrypt header without newlines and with any padding characters percent-encoded.
+ * `autocrypt-prefer-encrypt`: The prefer-encrypt value (either nopreference or mutual) of the address, as specified for the Autocrypt header. 
+
+The values derived from the mailto URI MUST not modify the existing Autocrypt Peer State for the given address. The values MAY only be used for composing the e-mail that was invoked by the mailto URI. If the local Autocrypt Peer State has a key for the given address and its `preliminary-recommendation` is calculated to be `available` (meaning not outdated and not received through gossip), that key MUST take precedent over the one supplied through the `mailto` URI.
```
missytake, line 1159: This could be formatted with https://sembr.org/ for better readability and better diffs?

doc/level1.rst
```diff
@@ -1142,6 +1142,27 @@ looking for a Setup Message again.  A more sophisticated
 implementation may offer a more user-friendly way to detect this
 situation and resolve it.
 
+Cold E-mail
+-----------
+
+tbd: this is why cold e-mail is cool
```
missytake, line 1148:
```suggestion
In-band key distribution has the disadvantage
that the first message between contacts is typically unencrypted,
as the sender has not yet the public key of the recipient.
But in many cases it is useful if users can initiate a conversation encrypted already.

One solution is the `Securejoin specification<https://securejoin.readthedocs.io/en/latest/>`_
which specifies a way of exchanging and verifying public keys
through a link which is exchanged through a trusted second channel.

If there is no direct channel available,
these are other ways of transferring the public key:

* Recipients can publish their identity on their public website with a :ref:`mailto: URI`.
* A mutual contact can relay the recipient's public key to the sender in a vCard file,
  next to other contact information.
```

And unless you already have something in mind for this section, feel free to take this :)

twiss:
Hello :wave: Apologies for the delayed response.
I would argue that the vCard topic is not specific to cold emailing; it can also be used for transferring your contacts+encryption statuses to a new device, among other use cases. So I would tend to put that in a separate document and then refer to that in the "Cold Email" topic among other places.

PR #456, opened 2025-06-16
Draft on importing PGP keys from vCards

missytake:
This documents the vCard format Delta Chat and Protonmail are emitting, and how they expect the key to be handled by an application.

Note that there are different use cases:
- recommending a contact to someone else
- sending your contact to someone so they can encrypt to you
- exporting contacts so you can import them on another device

They all carry different assumptions on the trust you can put on the key; so far we are conservative and always treat them as carrying little trust. This could be improved by distinguishing between different sources of keys.

- [ ] Add a menu item so the draft is visible in the sidebar on autocrypt.org
- [ ] Parse vCards with escaped comma in Protonmail


denzelonfire (approved):

doc/drafts/vcards.rst
```diff
@@ -0,0 +1,122 @@
...
+
+To update the Autocrypt Peer State, a vCard needs to contain at minimum:
+
+.. code-block:: vCard
+
+  BEGIN:VCARD
+  VERSION:4.0
+  FN:Dr. Alice Wise
+  EMAIL:alice@protonmail.com
+  KEY:data:application/pgp-keys;base64\,xlIEXd6t...ZFn/d2WPx4mOoF2ww0=
```
wiktor-k, line 85: Note that `application/pgp-keys` is for [ASCII-armored keys](https://www.ietf.org/rfc/rfc3156.html#section-7) and this seems to be using binary ones encoded using base64 so that MIME type is not appropriate IMO.

What is the correct one? I don't think there's any which fits the description. Maybe it's worth asking in the [OpenPGP Mailing List](https://mailarchive.ietf.org/arch/browse/openpgp/)?

missytake: To my understanding, we add the ascii-armored data + the armor checksum, as they are described in https://www.ietf.org/rfc/rfc2440#section-6.2, but without any newlines; I'm not sure if alternatives were discussed between others?

wiktor-k: https://www.ietf.org/rfc/rfc2440#section-6.2 lists other mandatory elements such as the header line. If you intend to use that then the `;base64,` part is wrong since it indicates *additional* encoding on top of the data format (so using that with OpenPGP ASCII-armored data would mean *double* base64).

Edit: two examples [from the Wikipedia](https://en.wikipedia.org/wiki/Data_URI_scheme#Syntax): `text/plain` doesn't need base64:

    data:text/plain;charset=UTF-8;page=21,the%20data:1234,5678

vs JPEG which is binary, and needs `base64`:

    data:image/jpeg;base64,/9j/4AAQSkZJRgABAgAAZABkAAD

andrewgdotcom: I think this is ok, we can specify (somewhere?) that “base64” in an openpgp context means base64 of the raw binary.

wiktor-k: > I think this is ok, we can specify (somewhere?) that “base64” in an openpgp context means base64 of the raw binary.

I think it'd require a note here: https://www.rfc-editor.org/rfc/rfc2397#section-2 alternatively the RFCs here need to be adjusted: https://www.iana.org/assignments/media-types/application/pgp-keys and the registration amended.

andrewgdotcom: I could add it to https://openpgp-media-types-73b9f4.gitlab.io

twiss: > To my understanding, we add the ascii-armored data + the armor checksum, as they are described in https://www.ietf.org/rfc/rfc2440#section-6.2, but without any newlines; I'm not sure if alternatives were discussed between others?

We put the binary keys there, which are then base64-encoded as per the Data URL. This ends up being very similar to an armored key without the header and footer (and without the checksum); but @wiktor-k you're right that we're technically violating RFC 3156.

In https://openpgp-media-types-73b9f4.gitlab.io/#name-new-media-type-parameters, @andrewgdotcom proposes to add an `encoding` parameter to the media type.

I'm not fully sure how that'd interect with a Data URL though, you'd have something like

```suggestion
  KEY:data:application/pgp-keys;encoding=binary;base64\,xlIEXd6t...ZFn/d2WPx4mOoF2ww0=
```

... and I'm not sure that's valid (or backwards compatible), but we could look into that.

(Also, in this context, `encoding=unarmored` would be more logical than `encoding=binary`, I think, since it's actually base64-encoded afterwards, after all.)

Alternative option would be to create a modifier and do something like

```suggestion
  KEY:data:application/pgp-keys+unarmored;base64\,xlIEXd6t...ZFn/d2WPx4mOoF2ww0=
```

(but same as before we may want to look into how existing applications handle that).

The simplest thing to do here would be to just specify that the default `encoding` is context-specific (or left unspecified).

wiktor-k: > I could add it to https://openpgp-media-types-73b9f4.gitlab.io/

Suggested snippet:

> when using `application/pgp-keys` with `base64` in the `data` URI (but not in all other cases) the ASCII-armor actually doesn't have the header and trailer lines and also no newlines contradicting RFC 2440

:kissing_cat:

andrewgdotcom: @twiss I see your point about base64-encoded "binary" 🤪 What about "plain" instead?

andrewgdotcom: Also, I'm not sure how the `;` separator in the content-type interacts with the `;` separator in the `KEY` record. Does one need to be escaped?

twiss: > Suggested snippet:
> 
> > when using application/pgp-keys with base64 in the data URI (but not in all other cases) the ASCII-armor actually doesn't have the header and trailer lines and also no newlines contradicting RFC 2440

This kind of re-frames a base64-encoded Data URL as containing some special form of armored key. I wouldn't do that; I would just simply state (in the context of this spec): "the Data URL contains an unarmored (binary) OpenPGP certificate, which is then base64-encoded as per the Data URL specification" (or some such).

> @twiss I see your point about base64-encoded "binary" 🤪 What about "plain" instead?

What's wrong with `unarmored`? I think it's the most explicit option, tbh. "plain" is very generic, so it's not really clear whether the parameter is specific to the OpenPGP encoding, or somehow about the encoding of the Data URL, or something else..

wiktor-k: > The simplest thing to do here would be to just specify that the default encoding is context-specific (or left unspecified).

IMO the simplest approach would be to define a MIME type for binary OpenPGP certs such as `application/openpgp-cert` and that's it :man_shrugging:

twiss: > Also, I'm not sure how the ; separator in the content-type interacts with the ; separator in the KEY record. Does one need to be escaped?

The ; separator comes from the Data URL spec; apparently it's possible to put media type parameters in there as per usual: https://www.ietf.org/rfc/rfc2397.txt. So they don't need to be escaped.

That being said, we'd need to check whether existing applications (including Proton Mail and Delta Chat) accept this without choking.

andrewgdotcom: In draft-ietf-openpgp-hkp I'm planning to define `application/pgp` for all binary PGP data, and leaving the existing ASCII-armored media types as effectively "legacy" types.

PR #461, opened 2026-03-28
Relax certificate structure requirements

Valodim:
Requiring a rigid certificate structure was intended to simplify compatibility for receivers of autocrypt headers, and ensure that clients normalize down to a minimal certificate before sending.

However, in practice:

1. this blocked adoption of Autocrypt for clients who wanted to use slightly different key structures
2. was not strictly followed anyways by several clients.

I believe the MUST should be a SHOULD there to resolve those two issues, while adding guidance on the intent of the structure.

This has also been requested by Kai here: https://lists.hostpoint.ch/archives/list/openpgp-email@enigmail.net/thread/SB56PWOWRURNX5MZI2UBWEC3OVS5WMEC/

### DIFF
--- a/doc/level1.rst
+++ b/doc/level1.rst
@@ -295,8 +295,8 @@ MUAs SHOULD treat such headers as invalid.
 OpenPGP Based key data
 ~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


kaie:
Thank you Vincent!

